PR #15, opened 2022-07-04
GHA config for PHP with fixed caching and testing for BC breaks

ciaranmcnulty:
Adds a matrix that installs the lowest-allowed versions of composer dependencies; this should detect any backwards compatibility breaks

Also fixes the composer caching (was pointing to wrong folder) and uses a slightly different 'set up PHP' action that is more widely used

### DIFF
--- a/.github/workflows/test-php.yml
+++ b/.github/workflows/test-php.yml
@@ -12,24 +12,42 @@ on:
 jobs:
   test-php:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php: ['8.1']
+        composer-mode: ['low-deps', 'high-deps']
 
     steps:
       - uses: actions/checkout@v3
 
-      - name: Cache Composer dependencies
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: "${{ matrix.php }}"
+
+      - name: Discover composer cache directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-dir)"
+        working-directory: php
+
+      - name: Cache composer
         uses: actions/cache@v3
         with:
-          path: /tmp/composer-cache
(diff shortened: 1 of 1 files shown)


.github/workflows/test-php.yml
```diff
@@ -12,24 +12,42 @@ on:
...
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-dir)"
+        working-directory: php
+
+      - name: Cache composer
+        uses: actions/cache@v3
         with:
-          working_dir: php
+          path: "${{ steps.composer-cache.outputs.dir }}"
+          key: composer
```
ciaranmcnulty, line 37: I've not done anything more complex here because composer doesn't evict unused stuff from the cache so AFAIK we'll just end up with a large cache of ZIPs of all the different versions of dependencies

.github/workflows/test-php.yml
```diff
@@ -12,24 +12,42 @@ on:
...
+        php: ['8.1']
+        composer-mode: ['low-deps', 'high-deps']
 
     steps:
       - uses: actions/checkout@v3
 
-      - name: Cache Composer dependencies
-        uses: actions/cache@v3
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
```
ciaranmcnulty, line 24: I don't know if this is 'better' than the other action but is sort of the default one OSS projects I'm involved with use

ciaranmcnulty:
So you don't want to merge or something?

aurelien-reeves:
> So you don't want to merge or something?

Yes, I do
It would just be nice to have the fix for "ramsey/uuid": "^4.2.1" (refs. https://github.com/cucumber/common/pull/2035/files#diff-0718aab1435f0377cfbb7f61cf52e015269dda9e98090b3ba9522f6612dbb489) to make the CI green

As this is a fork of you, it would be easier for you to push the patch

aurelien-reeves (approved):

PR #34, opened 2022-08-25
add workerId to TestCaseStarted

davidjgoss:
### 🤔 What's changed?

Add an optional workerId field to TestCaseStarted message.

See https://github.com/cucumber/cucumber-js/pull/2085#issuecomment-1184803712

Cucumber implementations that support parallel running via worker processes or some distributed model can set workerId when a test case is started.

Cucumber itself won’t do anything with this (at least initially), but it will allow third-party reporting tools to do some useful reporting on the parallel run.

### 📋 Checklist:

----


java/src/generated/java/io/cucumber/messages/types/TestCaseStarted.java
```diff
@@ -13,17 +13,20 @@ public final class TestCaseStarted {
...
     private final String id;
     private final String testCaseId;
+    private final String workerId;
     private final Timestamp timestamp;
 
     public TestCaseStarted(
         Long attempt,
         String id,
         String testCaseId,
+        String workerId,
```
mpkorstanje, line 23: Note that this makes it a breaking change.

davidjgoss: Oh because of ctor functions in some languages? Would it help to move it to the end?

mpkorstanje: Nope. Not for Java. But that's okay. As long as we follow Semver it's manageable.

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- BREAKING CHANGE: Add `workerId` field to TestCaseStarted message ([#34](https://github.com/cucumber/messages/pull/34))
```
mattwynne, line 12: Maybe we put this under the `### Changed` heading instead?

davidjgoss: Done

mattwynne (approved):

Issue #45, opened 2022-09-20
Release v19.1.3 (first release from this repo)


mattwynne:
I'm finishing the work on https://github.com/cucumber/common/issues/2029 and https://github.com/cucumber/messages/pull/27 by making an actual release. This ticket is where I'm going to document any issues I encounter as we test the release automation for the first time.

Releases:
- [x] Elixir / hex
- [x] Go
- [x] Java / Maven
- [x] NodeJS / NPM
- [x] C# / NuGet
- [x] PHP
- [x] Perl / CPAN
- [x] Ruby / RubyGems

mattwynne:
when I pushed to the release branch from my local, the push was rejected by GitHub because of two branch protection rules being violated:

1) linear history (there were apparently merge commits on `main`)
2) unsigned commits

I've disabled those checks for now (or at least, allowed admin override) as I think this might just be some cruft coming through from the monorepo, and we can re-enable it from now on.

mattwynne:
We need to change the go release to use the new `action-publish-subrepo` action

mattwynne:
PHP and Go are now working after I upgraded to v1.1.0 of the publish-subrepo action and removed a spurious `go` tag from those subrepos.

mattwynne:
Perl is the last one failing now.

We have this in the [error log](https://github.com/cucumber/messages/actions/runs/3100738583/jobs/5021342714):

```
...
Configuring Dist-Zilla-Plugin-Prereqs-FromCPANfile-0.08 ... OK
Building Dist-Zilla-Plugin-Prereqs-FromCPANfile-0.08 ... OK
Successfully installed Dist-Zilla-Plugin-Prereqs-FromCPANfile-0.08
26 distributions installed
Run filename=$(dzil build | tail -2 | head -1 | sed 's/.*writing archive to \(.*\)/\1/')
  filename=$(dzil build | tail -2 | head -1 | sed 's/.*writing archive to \(.*\)/\1/')
  echo "::set-output name=filename::$filename"
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
[GatherFile] /home/runner/work/messages/messages/perl/CHANGELOG.md does not exist! at inline delegation in Dist::Zilla::Plugin::GatherFile for logger->log_fatal (attribute declared in /usr/local/share/perl/5.30.0/Dist/Zilla/Role/Plugin.pm at line 67) line 18.
Error: Process completed with exit code 2.
```

[...]

PR #46, opened 2022-09-20
Fix elixir release

mattwynne:
Partial fix for #45

### DIFF
--- a/.github/workflows/release-hex.yaml
+++ b/.github/workflows/release-hex.yaml
@@ -13,10 +13,7 @@ jobs:
       contents: write
     steps:
       - uses: actions/checkout@v3
-      - uses: erlef/setup-beam@v1
-        with:
-          otp-version: 25.0.3
-          elixir-version: 1.13.4
+      - uses: ./test-elixir.yml
       - uses: cucumber/action-publish-hex@v1.0.0
         with:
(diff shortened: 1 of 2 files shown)


mattwynne:
I've tested these by pushing to `release/v19.1.3` you can see the progress [here](https://github.com/cucumber/messages/actions/workflows/release-hex.yaml?query=branch%3Arelease%2Fv19.1.3).

It looks like the last error is because I've forgotten about https://github.com/cucumber/polyglot-release/issues/80 so we're [still on v15](https://github.com/cucumber/messages/blob/main/elixir/mix.exs#L5).

Whoops.

mattwynne:
Instead of getting blocked by https://github.com/cucumber/polyglot-release/pull/82 I made a manual change to the `mix.exs` and the [release worked](https://github.com/cucumber/messages/actions/runs/3099428145)!

We'll have a small risk that this doesn't work the next time we release if `polyglot-release` isn't behaving exactly like my manual change, but we should be very close.

mattwynne:
Go is now working after I [fixed a bug in the publish-subrepo](https://github.com/cucumber/action-publish-subrepo/commit/e01f0b5fb1bdb10e8b2f549f07697a4b36518f4e) action.

PR #53, opened 2022-09-24
Fix jar check script in GHA

ciaranmcnulty:
The `scripts/java/check-jar.sh` was being run for the Java workflows, but silently failing because of a missing dependency on `xmlstarlet`

- Restricts the check to the ubuntu builds
- Explicitly installs xmlstarlet via apt
- Adds the `-e` flag to the script so similar errors in subshells will cause the main script to exit appropriately

We could also install the dependency in the Mac and Windows runs, if it's necessary. Although maybe checking the JAR in one context is enough?

### DIFF
--- a/.github/workflows/test-java.yml
+++ b/.github/workflows/test-java.yml
@@ -39,5 +39,9 @@ jobs:
       - run: mvn test
         working-directory: java
 
-      - run: ./scripts/check-jar.sh
+      - name: Check JAR
+        if: matrix.os == 'ubuntu-latest'
+        run: |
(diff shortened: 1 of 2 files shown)


mpkorstanje:
This script may also exist for cucumber-expresions. 

I think it can be removed though. We don't shade any depencies any more.

ciaranmcnulty:
@mpkorstanje is it useful to have locally or shall I just delete?

mpkorstanje:
No. Just delete it.

PR #55, opened 2022-09-24
Update PHP templates to have correct code style without reformatting

ciaranmcnulty:
Removes the use of php-cs-fixer to format the generated PHP, which means code generation can run without PHP

php-cs-fixer is used in the CI steps for PHP so we are still confident the code fits the ruleset


ciaranmcnulty:
I'd like to do this for the `go` version too but the ERB required is beyond me (it involves comment blocks lining up)

davidjgoss (approved):

mpkorstanje (approved):

PR #56, opened 2022-09-24
Combine the different test+release workflows, and put codegen check 1st

ciaranmcnulty:
These don't need to be separate files and keeping them together removes duplication of the trigger conditions


mpkorstanje:
They don't need to be, but for the sake of modularity and consistency, it would be better if they remain separated.

One of the biggest problems we currently have is ensuring all repos do everything the same way. Having individual files makes it easier to check this. For example we can compare the contents of all files with the same name and see that they see they are the same. With a single file this becomes much more complicated.

So while a single file may be locally optimal, across all projects it is sub optimal.

mpkorstanje:
For example I already see that the different actions use

```
    permissions:
      contents: write
```
inconsistently. We'd have to fix this in all projects at some point. This can be done with a little bit of scripting. But would need the structure across all projects to be the same.

mattwynne:
For context, the main reason we chose to use separate `release-*.yml` and `test-*.yml` scripts was that, at the time, GitHub did not let you re-run failed jobs, only failed workflows. So if you did a release and everything worked except the NPM release, you'd have had no choice but to re-run the Maven, RubyGems etc. releases as well if they were all in the same workflow. This is no longer true.

I'm not a big fan of big files, but I do really like the way GitHub visualizes the build for you when you express it like this. I suspect this is how we would have done it had GitHub had all of today's features when we started the project.

What worries me about having everything in one file like this is that it's much easier to miss a detail or inconsistency. I wonder if we could consider making re-usable workflows for the releases so that, if we go with this style, the yaml files will be much shorter and any variation will be easier to spot. We already have [one of these](https://github.com/cucumber/.github/blob/main/.github/workflows/prerelease-checks.yml) that's used [here](https://github.com/cucumber/cucumber-ruby-core/blob/9b3c892c4056240be6542be05a2df6b5062b68e9/.github/workflows/release.yml#L11) for example.

ciaranmcnulty:
One workflow can trigger another nowadays; possibly if we do that we'd get the visualisation?

mattwynne:
> For example I already see that the different actions use
> 
> ```
>     permissions:
>       contents: write
> ```
> 
> inconsistently. We'd have to fix this in all projects at some point. This can be done with a little bit of scripting. But would need the structure across all projects to be the same.

That's interesting! It's actually _easier_ to spot this when they're all together in the one file, at least for me. FWIW I think the only action that needs this permission is the GitHub release one.

I'd like to experiment with this. I agree that consistency is a good goal, but I don't want it to constrain us from innovation. We always have the option to undo the change in the future.

mpkorstanje:
> That's interesting! It's actually easier to spot this when they're all together in the one file, at least for me. 

You are mistaken about the point of this example. The point of the example was to show that we will have mistakes.

And fixing these across repos should be done programmatically. This will be easier if each language has its own file.

>  I agree that consistency is a good goal, but I don't want it to constrain us from innovation. We always have the option to undo the change in the future.

I don't see any innovation going on here. Code de-duplication isn't anything new and has well known tradeoffs.

In the future we'll still have the concerns about consistency, so I see no good reason to optimize against fixing those concerns in the present.

I'm currently thinking of creating another tool like polyglot release that (re-)generates these files on command.

mattwynne:
> > I agree that consistency is a good goal, but I don't want it to constrain us from innovation. We always have the option to undo the change in the future.
> 
> I don't see any innovation going on here. Code de-duplication isn't anything new and both keeping and fixing it have well known tradeoffs.

The innovation is that we're using GitHub Actions in a more idiomatic way, which for example means that the jobs' dependencies are visualised in GitHub's UI:

Screenshot 2022-09-29 at 5 49 50 PM

mattwynne:
> One workflow can trigger another nowadays; possibly if we do that we'd get the visualisation?

I don't think so.

mattwynne:
Rien and I just had a conversation about this. His point, and I think he's right, is that we are a small team working across a lot of repos, so consistency is really important. He has some ideas to develop (or use something like [this](https://github.com/todogroup/repolinter)) a tool to help check that these kinds of copy-pasted boilerplate are consistent across our repos. Introducing inconsistency in one place, even if we think it's an improvement, would make this harder.

He has nothing against the change per se, just that it's going to make things inconsistent.

@ciaranmcnulty if we were going to merge this, we'd need to commit to going through [the other repos](https://github.com/search?p=3&q=topic%3Apolyglot-release+org%3Acucumber+fork%3Atrue&type=Repositories) that use this test/release process and implementing the same change.

What do you think? Would you help me to get that done?

PR #58, opened 2022-09-25
Draft: Buildkit-based local builds

ciaranmcnulty:
## Background

@mattwynne and I paired on a spike to see what a Docker/Buidkit-based build process would be like

The main focus was local development, so although in this PR I've converted the linux GHA runs, that's mostly for information and has been isolated into a separate commit

This change lets a single command run all code generation, dependency installation, and testing. The only dependency is on a recent buildx, which is now included with Docker [^1]. There are no requirements for _any_ other shell commands to be present on host, including `bash` or `make`.

## Usage

- `docker buildx bake` generates the code, outputs it to disk, and runs the tests
- `docker buildx bake codegen` generates the code and outputs it to disk
- `docker buildx bake test` generates the code and then runs the tests without outputting it to disk

## Execution optimisation

Here's what will be executed when doing some typical developer workflows (after first run assumes you didn't clear buildkit cache):

 - ❌ not run
 - ✅ run for affected language
 - 🌍 run for all languages

| Scenario | Pull docker images | Code generation |  Dependency installation | Tests | |
| --- | --- | --- | --- | --- | --- |
| First run | 🌍 |  🌍 |  🌍 |  🌍 | Mostly bandwidth-limited  |
| No changes | ❌ |  ❌ |  ❌ |  ❌ | <10s  |
| Modified non-codegen code | ❌ |  ❌ |  ❌ | ✅ |   |
| Modified dependencies | ❌ |  ❌ |  ✅ | ✅ | Tests run if different packages are installed as a result  |
| Modified codegen template | ❌ |  🌍 |  ❌ | ✅ | Langs with same codegen output won't run tests |
| Modified JSON schema | ❌ |  🌍 |  ❌ | 🌍 |

### Cachebusting

The only part of the process that is not reproducible from local assets is the dependency installation. Because of this, each language's build has a stage named `with-dependencies`. 

This enables us to run `docker buildx bake --set '*.no-cache-filter=with-dependencies` to cachebust that stage only and force all the package managers to re-execute.

This could be useful in a nightly build, or a pre-release.

## Limitations

The way generated code is output to disk by Buildkit does not support file deletion. This means that for PHP and Java where the generated code is across multiple files, it may make sense to still have a small shell wrapper that deletes the files first.

 [^1]: There is also a way we can support arbitrarily old docker clients with a small shell script


ciaranmcnulty:
Ah we don't have GHA runners: see https://github.com/ciaranmcnulty/messages/pull/2 for the checks

mattwynne:
> Ah we don't have GHA runners: see [ciaranmcnulty#2](https://github.com/ciaranmcnulty/messages/pull/2) for the checks

Can you say more about this? Is there something I can change in the settings to get all the checks to run here?

mattwynne:
I like this. I think the only thing we're missing for me is a `CONTRIBUTING.md` which outlines how you use these commands for typical development tasks, what you need to install etc.

mattwynne:
The `CONTRIBUTING.md` added in https://github.com/cucumber/messages/pull/111 will need updated

mpkorstanje (approved):
This is very good looking. I hope we can do the same for Gherkin.

mpkorstanje:
@mattwynne you've merged this titled "Draft:" and you comment https://github.com/cucumber/messages/pull/58#issuecomment-1324632439 wasn't addressed. :D

PR #102, opened 2022-11-12
add global hooks

davidjgoss:

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

### 📋 Checklist:

  - [ ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../CHANGELOG.md), linking to this pull request.

----


davidjgoss:
A bit of a chicken and egg problem with the perl tests here - they rely on the current published version of the CCK, whose messages don't include the new field, which is causing the failure. Any thoughts on how to address this @ehuelsmann?

go/messages.go
```diff
@@ -283,9 +298,10 @@ type StepDefinitionPattern struct {
 }
 
 type TestCase struct {
-	Id        string      `json:"id"`
-	PickleId  string      `json:"pickleId"`
-	TestSteps []*TestStep `json:"testSteps"`
+	Id               string      `json:"id"`
+	TestRunStartedId string      `json:"testRunStartedId"`
```
mpkorstanje, line 302: This is a breaking change. 😃

php/src-generated/TestRunStarted.php
```diff
@@ -25,6 +25,7 @@ final class TestRunStarted implements JsonSerializable
      *
      */
     public function __construct(
+        public readonly string $id = '',
```
ciaranmcnulty, line 28: This is a breaking change due to the order of arguments. A call like this would previously have worked:

```php
$t = new TestRunStarted(new TimeStamp());
```
but now will not (because a timestamp is not a string)

For this reason new fields must be added to the end of the list if at all. I remember @mpkorstanje saying the issue is more complex for Java constructors

davidjgoss: Two things:

- Even appending to the end of the ctor is still a breaking change for Java
- It’s fundamentally a breaking change anyway because this new field is required, so I don’t think the order matters?

mpkorstanje: I think it's mostly a matter of putting the entry in the right place so we know to bump the major version.

ciaranmcnulty: I could be wrong but adding a required field breaks the ability to decode older messages - without explicit versioning in the streams that could be an issue

davidjgoss: You're right, in Java at least it will throw if a required field isn't present.

Would it then be better to introduce these fields as nominally optional? We could consider marking them required sometime in the future once they are consistently implemented.

mpkorstanje:
> A bit of a chicken and egg problem with the perl tests here - they rely on the current published version of the CCK, whose messages don't include the new field, which is causing the failure. Any thoughts on how to address this @ehuelsmann?

Looks like they're also not invoked for PHP:
```
There were 2 skipped tests:

1) AcceptanceTest::testAllNdJsonSurvivesDecodingThenEncoding
Test for AcceptanceTest::testAllNdJsonSurvivesDecodingThenEncoding skipped by data provider
PHPUnit\Framework\SkippedTestError: <no message>

2) AcceptanceTest::testAllFileStreamsSurviveDecodingThenEncoding
Test for AcceptanceTest::testAllFileStreamsSurviveDecodingThenEncoding skipped by data provider
PHPUnit\Framework\SkippedTestError: <no message>
```

php/src-generated/TestRunStarted.php
davidjgoss: ☝️ I've done this.

java/src/generated/java/io/cucumber/messages/types/TestCase.java
```diff
@@ -36,6 +39,10 @@ public java.util.List<TestStep> getTestSteps() {
         return testSteps;
     }
 
+    public Optional<String> getTestRunStartedId() {
+        return Optional.ofNullable(testRunStartedId);
```
mpkorstanje, line 43: What  is the rationale for making the field optional? I don't think there is a sensible behavior for a consumer that needs this field and doesn't get it.

davidjgoss: If we make it required now, the new code will be immediately incompatible with messages generated from the old code. This means e.g. we'd need to wait for all implementations to add this before we could use the new messages version in formatters.

If we're fine with that then I'm happy to change. But it seemed better to add it as optional, and maybe circle back and make it required down the line once it's established everywhere.

mpkorstanje: Okay, that's acceptable. Can you make a ticket so we won't forget to make it required?

davidjgoss: Done https://github.com/cucumber/messages/issues/113

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Added
 - BREAKING CHANGE: Add `id` property to `TestRunStarted`, optionally reference in `Attachment`, TestCase` and `TestRunFinished` ([#102](https://github.com/cucumber/messages/pull/102))
+- BREAKING CHANGE: Add `type` property to `Hook` ([#102](https://github.com/cucumber/messages/pull/102))
```
mpkorstanje, line 11: Shouldn't these go under the `### Changed` header?

mpkorstanje:
No thoughts yet. 

> So that would mean putting testRunStartedId on a TestCaseStarted message instead.

I think this would a more correct solution than adding it to the `TestCase`. Looking at the now out of date diagram, all the execution sits on the left hand side.

mattwynne:
> I originally added testRunStartedId to the TestCase message here. I've since removed that, having reconsidered. I think there's an argument that, given the (currently theoretical) scenario of multiple test runs in the same stream, a test case might be defined once and used in multiple test runs. So that would mean putting testRunStartedId on a TestCaseStarted message instead.

Is there another existing way to know which test cases are in which test run, before it starts? I'm thinking of a system to shard a large test suite across different execution nodes, where you'd have filtered the test cases into groups for distribution. In that situation you'd want to be able to associate the compiled test case with the test run. Getting that information up-front would mean you knew how many results you were waiting for, could show a progress bar etc.

philipphoertnagl:
Is there anything to change in the Running configuration or other setup for Cucumber Tests since this change?
In my Java Spring Boot project with v26.0.1 all runs fine, changing to v27.0.0 or v27.0.2 Cucumber tests won't start and I get an error:

`Exception in thread "main" java.lang.NoSuchMethodError: 'void io.cucumber.messages.types.TestRunStarted.<init>(io.cucumber.messages.types.Timestamp)'
	at io.cucumber.core.runtime.CucumberExecutionContext.emitTestRunStarted(CucumberExecutionContext.java:88)
	at io.cucumber.core.runtime.CucumberExecutionContext.startTestRun(CucumberExecutionContext.java:57)
	at io.cucumber.core.runtime.CucumberExecutionContext.runFeatures(CucumberExecutionContext.java:144)
	at io.cucumber.core.runtime.Runtime.run(Runtime.java:78)
	at io.cucumber.core.cli.Main.run(Main.java:87)
	at io.cucumber.core.cli.Main.main(Main.java:30)`

I use cucumber.version 7.20.1
Since this is a very new change I cannot find other people with this issue but I thought I could get help directly from you guys.
Thank you!

Issue #95, opened 2022-11-02
php published to archived sub repo.


mpkorstanje:
@mattwynne unlike cucumber-expressions and tag-expressions this project publishes to a sub repo. 

https://github.com/cucumber/messages/blob/b89d40438cf6b91fc4a246295054962354f86390/.github/workflows/release-go.yml#L7-L21

And while https://github.com/cucumber/messages-go does exist, I've archived it because it wasn't getting updated. I suppose that pushing to a sub repo is in error?

mpkorstanje:
Looks intentional to me. Resolved by unarchiving the repository's for go and php.

@mattwynne what about cucumber-expressions and tag-expressions?

mpkorstanje:
>> We currently publish a subtree of the go directory to https://github.com/cucumber/messages-go, is this still nessesary?
>
>I'm not sure, I think `github.com/cucumber/messages-go` was planned for deprecation (and maybe archiving), originally with the >idea of replacing it with monorepo, now with this repo. So probably any automations/maintenance for `messages-go` could be >disabled.
>
_Originally posted by @vearutop in https://github.com/cucumber/messages/issues/101#issuecomment-1312815858_

This means that:

We can remove the release-go workflows for 

- cucumber-expressions
- tag-expressions 
- messages
- gherkin

The release-github action and tags should be sufficient.

mattwynne:
I've discussed this with @mpkorstanje today. We've decided that:

* For Go, we'll stop using subrepo publishing
* For Go, we'll stop creating special tags https://github.com/cucumber/tag-expressions/blob/main/.github/workflows/release-go.yml#L24

For PHP, I'd like a final say-so from @ciaranmcnulty about the way forward. We're currently inconsistent - this repo is publishing to a https://github.com/cucumber/messages-php subrepo on release, but would it be OK to just use the main polyglot repo as we've decided to do for Go?

mpkorstanje:
This now also breaks on builds that are tagged with `go/vVersion`: https://github.com/cucumber/gherkin/actions/runs/3788087513/jobs/6440509357

ciaranmcnulty:
Now we've broken up the monorepo we certainly could publish from the polyglot repos (previously we could not have published multiple PHP packages from a single repo). 

There are a few factors against it, so it depends how they stack up against the neatness benefits:

[...]

PR #109, opened 2022-11-17
go: Remove release action

mpkorstanje:
It is sufficient to release go using only the release-github action


mpkorstanje:
Partial for https://github.com/cucumber/messages/issues/95

PR #118, opened 2022-11-30
Improve the Maven caching in docker builds build

ciaranmcnulty:
### 🤔 What's changed?

Fixes the commands used to pre-populate the maven local repository. This was wrongly doing `mvn install` to prime the dependencies, but that does something quite different

We don't want to be running the tests during the 'install dependencies' phase

`mvn verify` at the end still pulls in some new dependencies every time - how can we prevent this?

### DIFF
--- a/java/Dockerfile
+++ b/java/Dockerfile
@@ -4,7 +4,7 @@
 ARG MAVEN_VERSION=3.8
 ARG JAVA_VERSION=16
 FROM maven:${MAVEN_VERSION}-eclipse-temurin-${JAVA_VERSION} AS java
-
+WORKDIR /cucumber
 
 # Dummy stage for generated code, overriden in main build
 FROM scratch AS schema-codegen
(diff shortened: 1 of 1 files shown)


java/Dockerfile
```diff
@@ -4,16 +4,17 @@
...
 FROM scratch AS schema-codegen
 
 
 FROM java AS with-dependencies
 
-COPY --link pom.xml maven-versions-rules.xml .
-RUN mvn install
+COPY --link pom.xml .
+# verify loads additional dependencies without running the tests (they aren't copied in yet)
+RUN mvn dependency:go-offline verify
```
mpkorstanje, line 17: This `verify` command will run test. I don't think it's needed here

ciaranmcnulty: It triggers some extra dips to load but pas per the comment because we haven't copied the sources no tests run

mpkorstanje: Aan gotcha! 😄

mpkorstanje (approved):

PR #122, opened 2022-12-11
Add exception to TestStepFinished TestRunFinished

mpkorstanje:
### 🤔 What's changed?

To implement a message based JUnit XML Formatter (https://github.com/cucumber/cucumber-junit-xml-formatter/pull/3) we need to know the class and message of the exception that caused the scenario to fail. This information is often included implicitly in the `message` field but can not be reliably extracted.

For example if a scenario failed duo to a assertion exception the failure should be rendered as:

```xml
<failure message="[The scenario has pending or undefined step(s)](expected: <\"a\"> but was: <\"b\">)" type="org.opentest4j.AssertionFailedError">
  <![CDATA[

... Detailed stack trace here

]]>
</failure>
```

Currently however we can only render the stack trace part. The other information is not available.

Is this generic enough to work in other languages?

### 📋 Checklist:


davidjgoss:
> Is this generic enough to work in other languages?

I think it's okay at least from a JS persective. We can always provide a generic fallback if those fields are not present?

You mentioned the flatness of the structure before, I guess the alternative would be an `exception` sub-object with `type` and `message` fields - this would then be a natural place for additional granular detail about the exception to go e.g. actual vs expected when we eventually get to that.

mpkorstanje:
Good one. There are some proposals to add more information about exceptions.

- https://github.com/cucumber/messages/issues/70
- https://github.com/cucumber/messages/issues/69

mpkorstanje:
I've added an Exception object with `type` and `message` fields. I looked at adding `stacktrace` and `expected` and `actual` but it's a bit too much work to see that every language can do with stack traces.

davidjgoss:
What's the use case for the exception on `TestRunFinished` out of interest?

mpkorstanje:
To inform about exceptions that didn't happen in steps or global hooks.  For example in Cucumber JVM this would be related to creating the dependency injection container.

Cucumbers execution model can be seen as a tree of containers. Each container has a started and finished event. Each finished event should have an exception for exceptions that occurred in that container.

PR #135, opened 2023-01-06
Add explicit CCK download to make PHP acceptance tests run

### LINKED ISSUE #134: PHP version doesn't run acceptance tests since monorepo split
The PHP implementation used to look at the compatibility-kit folder to get the ndjson example files.

Because of the way this was implemented, a non-existent folder = no source data = skipped tests :/

The Ruby version also uses the CCK to test, but it looks like CCK is published as a bundle and data's sourced from there - the PHP implementation should  explicitly download the CCK data and use it for the tests.

ciaranmcnulty:
Fixes #134 

End to end tests were skipped because the path to CCK sample data was broken during extraction from the monorepo. 

 - adds a separation between unit tests and the (one) end to end test
 - makes the end to end test fail loudly if the CCK samples aren't present
 - explicitly downloads the CCK samples from GitHub during the docker build


luke-hill:
@ciaranmcnulty - We're releasing a major version of the CCK (v13), soon.

95% of it is ruby breaking changes, but there is one global change (2x bugfixes to attachment).

If you want to follow it: Progress is here: https://github.com/cucumber/compatibility-kit/pull/52

luke-hill:
I've had a thought about this recently, and wouldn't it be better to just have php conformance inside the CCK proper? You could even slowly introduce it for say 1 sample at a time if there are things which aren't quite right.

mpkorstanje:
@luke-hill dependency wise that doesn't work. Using the cck to test messages creates a circular dependency. 

And do keep in mind that this pull request is there to make sure tests are ran at all. It is not the perfect solution, which would be to pare down the tests in messages just to the point where they test if serialization works. Because all messages are generated, a few well chosen tests can validate that all messages work.

If possible, consider merging this as and then we can improve from there.

PR #152, opened 2023-05-03
Feature/cpp

chybz:
### 🤔 What's changed?

- Added C++ message generation support
- Added CMake library

While adding C++ support in [Gherkin](https://github.com/chybz/gherkin/tree/feature/cpp), saw that other implementations were using this repository, so decided to contribute in here as well.

### 📋 Checklist:

----


cpp/src/lib/cucumber-messages/CMakeLists.txt
```diff
@@ -0,0 +1,142 @@
+add_library(cucumber-messages_library)
+add_library(cucumber::messages ALIAS cucumber-messages_library)
+
+set(INC_DIR "${CMAKE_SOURCE_DIR}/include")
+
+target_sources(
+    cucumber-messages_library
+    PRIVATE
+        ${INC_DIR}/cucumber/messages/attachment_content_encoding.hpp
```
mpkorstanje, line 9: It might be prudent to generate this list as well. Otherwise adding a new message may break the CPP project.

mpkorstanje (commented):
Cheers. 

> No external dependencies, so no JSON support as of now, hence no tests.

I think we can do without tests for serialization, but I would expect at the very least a `test-cpp.yml` in the github workflows to ensure the project compiles.

> Arbitrary dependency on C++20

I'm not familiar with CPP.  Where does the dependency on C++20 come in to play?

cpp/src/lib/cucumber-messages/CMakeLists.txt
chybz: Agreed. While it's considered an "anti-pattern" in CMake, I've always preferred it that way.

Changed accordingly.

As for the C++20 dependency, it requires a fairly recent compiler and may fail in some environments. Only a large and diverse user base could tell if this is a problem (enhanced language constructs, features and standard library).
It's reasonable to expect C++20 nowadays, but I cannot speak for other human beings...

mpkorstanje: You're currently the only active C++ contributor.  I think we can use you as representative sample of the userbase. 😄

jenisys:
> Arbitrary dependency on C++20

You should not require the C++20 standard in the CMake file, 
if you source code does not require some of the features from the C++20 standard. 

POSSIBLE APPROACHES:

* If you define `CMAKE_CXX_STANDARD` as default if no C++ standard is defined.
  You should support that the user overrides it via `cmake -DCMAKE_CXX_STANDARD=17 ...` on the command line.
  In addition, this approach allows that a parent `CMakeLists.txt` file (that wants to use this library as part of the overall build process) predefines the `CMAKE_CXX_STANDARD`.
* BETTER APPROACH: Your library(s)/program(s) should specify which C++ features are required.
  This determines which C++ standard is used (or it fails if the C++ compiler does not support these standard(s)/features)
* The C++ compiler version (as part of the toolchain) defines the default C++ standard that is used by default

```cmake
# -- FILE: CMakeLists.txt
# EXAMPLE FOR SOLUTION 1
if(NOT DEFINED CMAKE_CXX_STANDARD)
    set(CMAKE_CXX_STANDARD 20)  # Use C++20 standard as default
endif()
```

chybz:
Yes, definitely. That was just me being lazy. Thank you very much !

Issue #162, opened 2023-06-27
Add support of Python


elchupanebrej:
### 🤔 What's the problem you're trying to solve?

Python is a very popular language, but there is no reference implementation of messages for python

### ✨ What's your proposed solution?
Take out https://github.com/elchupanebrej/pytest-bdd-ng/blob/default/src/pytest_bdd/model/messages.py model as reference implementation of messages for python

mpkorstanje:
This projects generates all the code based on the JSON schema. 

New language implementations can be added adding a template to https://github.com/cucumber/messages/tree/main/jsonschema/scripts/templates and copying the `Makefile` from an existing language implementation.

Probably best to hop onto Slack if you're interested in contributing one.

luke-hill:
@elchupanebrej this is now happily closable 🎉

PR #165, opened 2023-07-18
Add messages implementation for python

elchupanebrej:
### 🤔 What's changed?

Add python implementation

### 📋 Checklist:

----


elchupanebrej:
This address to https://github.com/cucumber/messages/issues/162

elchupanebrej:
This was generated by https://github.com/koxudaxi/datamodel-code-generator with a small manual changes

mpkorstanje (changes requested):
At a glance this doesn't follow the pattern used by the other language implementations in quite a few ways.

I also don't understand the purpose of the samples directory.

mpkorstanje:
> For Python exists a tool that allows generating Pydantic models directly from json schema 

You can use Pydantic if you can make it fit into the `make clean-all generate-all` workflow. Though I suspect your manual edits might pose a problem.

> Samples are taken from gherkin repository to validate if serialization/deserialization works well.

Consider narrowing this down to a few representative examples.

python/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+description = "Cucumber Messages is a message protocol for representing results and other information from Cucumber. "
+license = {text = "MIT"}
+maintainers = [
+  {name = "Konstantin Goloveshko", email = "kostya.goloveshko@gmail.com"}
+]
+name = "messages"
+readme = {file = "README.md", content-type = "text/markdown"}
+requires-python = ">=3.8"
+urls = {Repository = "https://github.com/cucumber/messages"}
+version = "0.1.0"
```
luke-hill, line 37: This will need editing to conform to polyglot-release

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,59 @@
...
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
+    strategy:
+      matrix:
+        python-version:
+          - "3.8"
+          - "3.9"
+          - "3.10"
+          - "3.11"
+          - "pypy3.8"
```
luke-hill, line 26: Is this similar to 3.8? Your matrix is gonna get rather large rather quickly

luke-hill (changes requested):
If you're going to copy lots of the cck it would be better to fetch the data using some form of call rather than C+P as this is currently being rapidly updated

.github/workflows/test-python.yml
elchupanebrej: It won't be - the suite of tests is pretty small.

python/pyproject.toml
elchupanebrej: Could you please give a link where I could find how to do this right

elchupanebrej:
> At a glance this doesn't follow the pattern used by the other language implementations in quite a few ways. Please follow up the directions from #162 around code generation.
> 
> I also don't understand the purpose of the samples directory.

Agree, let do this in a few steps. At first - implementation, after that - generation

elchupanebrej:
> Samples are taken from gherkin repository to validate if serialization/deserialization works well.

> Consider narrowing this down to a few representative examples. Currently it is hard to see the forest for the trees.

Samples are used in tests. More complex tests could exist. I insist to include them for now

python/pyproject.toml
luke-hill: I don't think we directly have a link. But we use a tool called `polyglot-release` to release internal tools.

If you look at how it is used. You'll see you'll need a version file that is directly auto-edited.

https://github.com/cucumber/polyglot-release

An example of "how" it auto-worked from the CLI can be viewed in this commit for gherkin
https://github.com/cucumber/gherkin/commit/650109d73b70e83fe6c1eb400005084fd5a2f59d

Here is "what" I did.

Ran a command on the CLI, then waited a few seconds or so.
It auto-generated this commit and changed all the relevant files (Check the python change)

python/RELEASING.md
```diff
@@ -0,0 +1,15 @@
+# How to make a release
```
luke-hill, line 1: This here should be using the auto-copy from polyglot-release.

See an example in one of the ruby projects I manage here: https://github.com/cucumber/cucumber-ruby/blob/main/RELEASING.md

Which links to: https://github.com/cucumber/.github/blob/main/RELEASING.md

This will tell you how to get a few things configured and setup, specifically for some of the internal toolings.

If you need more dedicated help, join us on slack @ cucumberBDD.slack.com

mpkorstanje:
@elchupanebrej  

> Samples are used in tests. More complex tests could exist. I insist to include them for now

What purpose do they serve? They'll be a hassle to update if/when the schema changes.

luke-hill:
Hi @elchupanebrej - Just checking in to see where you're up to with this. Is this something you're still working on?

elchupanebrej:
> Hi @elchupanebrej - Just checking in to see where you're up to with this. Is this something you're still working on?

Hi @luke-hill, sorry for the long response, hadn't time to work on the project. I'll try to create another merge request that will conform to the building process.

elchupanebrej:
The PR was updated with Makefile. Model is stable, so generated code is totally same to version, which was generated at first try

@mpkorstanje I kindly ask you to review the code and take a release part. I didn't get into all deps&relations between release tools.

python/tests/test_model_load.py
```diff
@@ -0,0 +1,52 @@
...
+from pytest import fixture
+
+from _messages import Envelope
+
+
+@fixture
+def compatibility_kit_repo(tmpdir):
+    repo_path = Path(tmpdir) / "compatibility-kit"
+    repo = Repo.clone_from(
+        "https://github.com/cucumber/compatibility-kit.git",
```
mpkorstanje, line 16: Messages should not use the compatibility kit as this creates a circular dependency. Rather you'll want to write some targeted tests for serialization and deserialisation.

python/pyproject.toml
mpkorstanje: @luke-hill this currently looks good. Do you see a specific problem?

python/tests/test_model_load.py
mpkorstanje: @luke-hill the above comment also applies to you.

mpkorstanje:
Left a few quick remarks, will have to take a deeper look later.

python/Makefile
```diff
@@ -0,0 +1,32 @@
...
+.DEFAULT_GOAL = help
+
+MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+HERE := $(dir $(MKFILE_PATH))
+
+help: ## Show this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make <target>\n\nWhere <target> is one of:\n"} /^[$$()% a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+generate:
+	datamodel-codegen \
```
mpkorstanje, line 12: This will not work. But CI passed because the Makefile in root does not invoke code gen Python.

Ideally python uses the same approach as the other languages. This ensures that the code can be generated without installing every language.

Further more the generated models should be checked in. This ensures that all code changes can be inspected and that the code gen process is not required to build a specific language.

python/tests/test_model_load.py
elchupanebrej: Thanks! The small suite of tests will be copied here.
Java tests seem must not be directly ported because the model is generated from schemas directly. So many tests will just test the generator itself(it has a much wider suite of tests)

mpkorstanje: Indeed. Most of the tests in Java are for serialization rather than the shape of the messages.

For example enums must be serialized by name,  null fields must be omitted, ect. This will depend a bit on what Python offers out of the box.

python/Makefile
elchupanebrej: Other language realizations use additional tools. For Python such a tool is `datamodel-codegen` that needs Python environment (So a language interpreter is needed here not for generation itself but for launching a tool. ) . This tool targets the most specific task - generating Python models from json schemas. I insist on not competing with specific tools with own just written solution.

python/pyproject.toml
elchupanebrej: @luke-hill I did both solutions: copying cck to this repo directly and getting them from the repo at the test runtime. @mpkorstanje has asked to not depend on the cck repo itself, so there is just a small amount of tests

python/Makefile
mpkorstanje: Okay that's fair. I'll take a look at the build process. Perhaps we can fit in as is.

python/tests/test_model_load.py
luke-hill: It's worth pointing out here @elchupanebrej the way in which Rien is describing things is that we should / can use the CCK to test the generation. But we shouldn't have the generation "requiring on" the CCK. Hope that makes sense / apologies if I'm repeating something already understood.

i.e. for ruby here - https://github.com/cucumber/messages/blob/main/ruby/cucumber-messages.gemspec we have no direct dependencies, but we use the CCK as a development dependency (I.e. to test the generation has worked).

Apologies if this doesn't make sense.

python/pyproject.toml
luke-hill: @mpkorstanje this version should be 25+ not 0.1 - Although maybe we keep them async until we're happy it's stable and bump

python/src/message_samples/minimal/minimal.feature.ndjson
```diff
@@ -0,0 +1,12 @@
+{"meta":{"ci":{"buildNumber":"154666429","git":{"remote":"https://github.com/cucumber-ltd/shouty.rb.git","revision":"99684bcacf01d95875834d87903dcb072306c9ad"},"name":"GitHub Actions","url":"https://github.com/cucumber-ltd/shouty.rb/actions/runs/154666429"},"cpu":{"name":"x64"},"implementation":{"name":"fake-cucumber","version":"16.3.0"},"os":{"name":"darwin","version":"22.4.0"},"protocolVersion":"22.0.0","runtime":{"name":"node.js","version":"19.7.0"}}}
```
luke-hill, line 1: I think this is a good process what you've done here. Just commenting for documentation.

I think as/when you have gotten this all working, it would be good to migrate this and others to the CCK proper. WDYT? (Maybe something for 2025?)

python/pyproject.toml
elchupanebrej: I have added a release GitHub flow. If I understood correctly, would the version be up for release?

python/tests/test_model_load.py
elchupanebrej: PyPi releases don't allow dependencies on github repositories, so I can't  add resources directly. If you go through commits you will see an example of tests with direct downloading CCK data. If you have better ideas how to integrate - please share your thoughts

python/Makefile
elchupanebrej: Added dependency on python so genenerate-all flow works.

python/src/message_samples/minimal/minimal.feature.ndjson
elchupanebrej: > migrate this and others to the CCK proper
It must work with CCK now in all possible cases. If it doesn't - let write tests & fix

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,50 @@
...
+          - "3.10"
+          - "3.11"
+          - "3.12"
+          - "pypy3.8"
+          - "pypy3.9"
+          - "pypy3.10"
+        os:
+          - ubuntu-latest
+          - windows-latest
+          - macos-latest
```
mpkorstanje, line 33: While our CI is free, it is not unlimited and we are frequently running out of runners. So I think the current matrix is a bit excessive. 

Can we reduce this to all currently supported python versions on `ubuntu-latest` and then only the latest python version on `windows-latest` and `macos-latest`?

python/Makefile
mpkorstanje: As discussed directly. You can stub out the make file and invoke the code generation directly from the build process.

python/pyproject.toml
mpkorstanje: @luke-hill then the current version is fine. `polyglot-release` will replace the version number before release.

@elchupanebrej the first release with python in it will probably be a v26.x release. All languages are released in tandem.

python/src/message_samples/minimal/minimal.feature.ndjson
mpkorstanje: @elchupanebrej As a test I'm not happy with a "sample test". As said before this create a circular dependency between the code that generates the samples and messages.

Tests for messages can be limited to testing whether the code was generated and serialization works correctly. This is does not test those things specifically while still testing many other  - less relevant things.


@luke-hill what exactly do you mean by "migrating this and others to the cck"?

python/src/message_samples/minimal/minimal.feature.ts
```diff
@@ -0,0 +1,6 @@
+import assert from 'assert'
+import { Given } from '@cucumber/fake-cucumber'
+
+Given('I have {int} cukes in my belly', function (cukeCount: number) {
+  assert(cukeCount)
```
mpkorstanje, line 5: This file seems unused in any tests.

python/src/messages.py
```diff
@@ -0,0 +1,3 @@
+from _messages import *
+
+ExpressionType = Type1
```
mpkorstanje, line 3: I don't understand what this file does. Can you explain?

mpkorstanje (changes requested):
There seems to have been a misunderstanding. 

So just to clarify. 

Either:

* Source is generated by the Ruby codegen script 
* Generated source is checked in

Or:

* Source is generated by the python build process.
* Generated source is not checked in.
* Make targets print a message that code code gen is handled by Python.

Which option are you going for now?

python/src/_messages.py
```diff
@@ -0,0 +1,755 @@
...
+
+
+class MediaType(Enum):
+    text_x_cucumber_gherkin_plain = "text/x.cucumber.gherkin+plain"
+    text_x_cucumber_gherkin_markdown = "text/x.cucumber.gherkin+markdown"
+
+
+class Source(BaseModel):
+    model_config = ConfigDict(
+        extra="forbid",
```
mpkorstanje, line 24: If a field gets added, will this prevent an application using an older version of messages from de-serializing a message created by an application using the newer version of messages?

jsa34:
@elchupanebrej  I'm just trying this out locally to see, and noticed that pytest is missing from the dev dependencies in pyproject.toml

python/src/message_samples/__init__.py
jsa34, line 1: Is this supposed to be in the app's src directory?
These look like test data, so feel more at home in the /tests directory?

python/pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.8",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12"
+]
+dependencies = [
+  "importlib_resources",
+  "pydantic>=2.0.3"
```
youtux, line 27: Is it really necessary to use and add  pydantic as a dependency ? 
Many people are still on pydantic v1, and this would require pytest-bdd users to upgrade to pydantic v2 since pytest-bdd will soon depend on gherkin

Aren’t stdlib dataclasses enough?

jsa34: importlib_resources is also, from what I can see, only used for tests which I'm not sure is needed either

python/tests/test_model_load.py
jsa34: Surely this isn't a problem if it's listed as a dev dependency as it won't be in the pypi package?

elchupanebrej:
Thanks for great review, return later this week and will update all things accordingly 😀

python/tests/test_model_load.py
elchupanebrej: @luke-hill, this implementation isn't dependent on CCK for generation, it was dependent on CCK for test purposes only

python/src/message_samples/minimal/minimal.feature.ndjson
elchupanebrej: @mpkorstanje sorry for bothering you, it seems I can't catch a point:

Samples of messages in the CCK repository are stored as examples. Every tool that uses messages has to use them (at least serialize when some event is emitted, and deserialize when this message comes to some reporter). So I took the full suite of test data from the CCK repo and checked that the models generated were successfully parsed that messages into the model, and after that deserialized them to totally the same JSON. Could you please describe more precisely what kind of tests would be OK: would be enough if some model(for every kind of message) would be created, serialized and deserialized perfectly to the totally same model?

python/pyproject.toml
elchupanebrej: @youtux Yes, this is technically possible, but such realization will be dependent on some library like https://github.com/lidatong/dataclasses-json (the best option for now), which are not as good supported as pydantic

From another perspective - testing utilities are selected at the start of a project, so if the messages package will be used somewhere - it most probably would be dependent on the new version of Pydantic

python/src/messages.py
elchupanebrej: We have two entities in the original model named Type (design bug from my perspective). This module is a simple adapter, so the end user will import Type and ExpressionType but not Type and Type1. In the serialized model they both are named Type as it was in the original model

python/src/_messages.py
elchupanebrej: Reworked for forward compatibility

python/src/message_samples/__init__.py
elchupanebrej: Moved to tests/data folder

python/Makefile
elchupanebrej: Done

python/pyproject.toml
youtux: but there are many projects using pytest-bdd for years, and this would be an issue.
We can do without pydantic in a very simple way. We can use data classes, then when we need to serialise to json we call `asdict(model)`. If we need custom encoders (e.g. for date times) we can implement a simple JsONEncoder and pass that to `json.dumps(asdict(model), encoder=…)`.

Or also just implement custom serialiser for each object

elchupanebrej: in this case, we have to implement dict_factory for dataclass.asdict, which will have to take in count Enums, or there would be an issue with serialization to JSON. And deserialisation to the dataclass also will be an issue (Enums again)
And pydantic covers both of this issues

mpkorstanje: Minimizing the number of dependencies avoids a potential conflict with the system under test. And it seems to me that any effort saved by using Pydantic in Cucumber will be meaningless if Cucumber can't be used because of it. 

But I'm not in the Python ecosystem so I'd like to see a consensus on this problem from those who are.

mpkorstanje (changes requested):
I'm sorry to see that the misunderstanding I highlighted in the last review persists. While some aspects have been addressed, they have not been addressed in full. Let me know if we need to schedule a call and talk this through.  

Further more, the current test set contains many incidental details while also not testing for anything specific. This will make the tests break whenever small changes to the schema are made. Given that this repository currently hosts 9 languages, keeping tests up to date becomes tedious quickly.

Finally I'd like to see a consensus on the use of pydantic.

python/pyproject.toml
youtux: I really think we should not bring in a big dependency like pydantic here, especially since it has made a big API change in v2, and I can see it make it difficult for users to adopt this library if it conflicts with their pydantic v1 requirement.

What's the use of pydantic here? I don't see it being used for serialisation / deserialisation here.
What's the API of this library going to look like?

elchupanebrej: Messages library is hardly used for serialization/deserialization, for example:
- Test runner must produce messages in the ndjson format, so it uses model of "messages" lib to represent outcomes, messages lib serializes and validates against Json schema (non-directly).
- Test reporter consumes ndjson stream of messages and uses "messages" library to deserialize inputs and validate them.

So "messages" lib is a bridge between test runner and test reporter (potentially from different languages ecosystems)

youtux: ok, but how is the API of this lib supposed to look like?

```
from cucumber_messages import ???

???
```

elchupanebrej: @youtux , please check python/tests/test_model_load.py test in this PR (I'll rework tests later).

For example reporting in the pytest-bdd-ng uses this particular model:
https://github.com/elchupanebrej/pytest-bdd-ng/blob/default/src/pytest_bdd/message_plugin.py

jsa34:
Hello @elchupanebrej !

I wondered if you had had a chance to review this?

😃

youtux:
> I'm sorry to see that the misunderstanding I highlighted in the last review persists. While some aspects have been addressed, they have not been addressed in full. Let me know if we need to schedule a call and talk this through.
> 
> Further more, the current test set contains many incidental details while also not testing for anything specific. This will make the tests break whenever small changes to the schema are made. Given that this repository currently hosts 9 languages, keeping tests up to date becomes tedious quickly.
> 
> Finally I'd like to see a consensus on the use of pydantic. And it might be useful to do that first as it will significantly impact the shape of this pull request.

To me it's still not quite clear how this implementation (but the others already present in the repo as well) is supposed to be used.
This repo should just define the json schema for the Cucumber Messages specification, but it has implementations, and it doesn't show anywhere how these implementations are supposed to be used.
I just [asked the same question on Discord](https://discord.com/channels/1260995505018765393/1264635611994652744/1299471209641148479), maybe they will help with the answer.

youtux:
If we are doing this implementations just to define the classes, I think we should better use an automated tool that converts from jsonschema to python models.

A good one seems to be https://github.com/koxudaxi/datamodel-code-generator, I managed to create dataclasses from the jsonspec like this:
```
 docker run --rm -v "${PWD}:/local" koxudaxi/datamodel-code-generator --input /local/jsonschema/GherkinDocument.json  --output /local/model.py --output-model-type dataclasses.dataclass
```

It also supports pydantic v1/v2, and other libs, if one needs that.
A this point, I guess we should just create a README.md that explains users that want to have python classes from the JSONSchema how to do that with this tool.

youtux:
If anything, we could maintain the generated files into this repo, and make sure every time the json schema is updated these files are regenerated, so that other downstream users can use whatever flavour of models they want.

elchupanebrej:
> we should better use an automated tool that converts from jsonschema to python models.

@youtux, this approach and tool are used here exactly! Check Makefile pls!

python/src/message_samples/minimal/minimal.feature.ndjson
mpkorstanje: The CCK uses the messages to generate the output of a canonical cucumber execution. For this is needs the messages. The value the CCK adds isn't that it generates a sample of each messages, but rather that the collection of messages as a whole. So it can for example express relationships between messages.

This dependency also means that it can't be used as test data.

Now for messages the exact testing strategy depends on the framework and language used.

For example for Javascript, the object and it's json representation are almost identical so there is little to test at all. And because the code is generated, it doesn't seem nesesary to test every message either.

So you can see we do a round trip test of one moderately complex message and not much more.

https://github.com/cucumber/messages/blob/main/javascript/test/messagesTest.ts

For Java serialization is more complicated. It does not have a concept of `undefined`. So we got tests to check for that.

https://github.com/cucumber/messages/blob/main/java/src/test/java/io/cucumber/messages/NdjsonSerializationTest.java

Now I don't know enough about Python to tell you exactly what to test. I can't tell you about pitfalls I don't know about. But I imagine if third party code generator is used, a simple round trip should be enough.

mpkorstanje:
> To me it's still not quite clear how this implementation (but the others already present in the repo as well) is supposed to be used.

A schema definition isn't any good without data objects to go with it and their use is mostly to provide a type safe representation of the message. 

For example 

https://github.com/cucumber/junit-xml-formatter/blob/main/java/src/main/java/io/cucumber/junitxmlformatter/XmlReportWriter.java

And while in theory each library could generate dtos based of the schema, that isn't practical once libraries start calling each other. So a shared implementation of the data objects is essential.

youtux:
Got it. Then I'd propose for the python impl to provide at least the data classes versions, since its the most compatible one, and possibly also the pydantic version under a different module, so that downstream users can choose what to use

jsa34:
Is there anything I can help with here, as I'd love to try and messages over the line to support our work with gherkin?

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,87 @@
...
+          - python-version: "pypy3.10"
+            os: windows-latest
+          - python-version: "3.8"
+            os: windows-latest
+          - python-version: "3.9"
+            os: windows-latest
+          - python-version: "3.10"
+            os: windows-latest
+          - python-version: "3.11"
+            os: windows-latest
```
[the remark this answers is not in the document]
luke-hill: I would use an include matrix personally @elchupanebrej

luke-hill (commented):
I'm going to step away from this now I think as there is enough voices commenting on it. I don't quite understand why we've not gone down the route that the other 8 languages have done with using the codegen tool given I spent a while refactoring it so now it's a tiny class you need to make :shrug:

python/src/message_samples/minimal/minimal.feature.ndjson
luke-hill: To add as well. For the ruby implementation we don't even do something that complex. All we do is "open" the message (Which is an `Envelope` class), then reclose it again. And check that in doing so we don't change anything.

If you're finding that you're going down the rabbit hole of doing lots of testing here, then something is likely going wrong. Messages are the building block that everything else is based off, not the other way around.

python/.pre-commit-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+  - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+    rev: v2.14.0
+    hooks:
+      - id: pretty-format-toml
+        args: [--autofix]
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.17.0
+    hooks:
+      - id: pyupgrade
+        args: ["--py38-plus"]
```
jsa34, line 34: Should this not be 3.9 as opposed to 3.8?  (Looking at pyproject.toml)

elchupanebrej: Thanks! Done!

python/src/messages.py
elchupanebrej: Done

python/src/cucumber_messages/json_converter.py
```diff
@@ -0,0 +1,321 @@
+import json
+import re
+import sys
+import types
+import typing
+from dataclasses import MISSING, Field, fields, is_dataclass
+from datetime import date, datetime
+from enum import Enum
```
jsa34, line 8: Considering we are targetting python >=3.9, I was surprised that we are using typing imports for things like Dict, List, Union, etc. instead of the built-ins?

elchupanebrej:
This would use https://github.com/lidatong/dataclasses-json/issues/442 in future

elchupanebrej:
@mpkorstanje @luke-hill @jsa34 @youtux 
Please make a new review for this PR:

1. Model doesn't use Pydantic anymore (dataclasses are used)
2. Model generation leverage on ruby templates (so no more Makefile workarounds)
3. Tests are provided in the same manner as at other language implementations

python/src/message_samples/minimal/minimal.feature.ndjson
elchupanebrej: Copied ruby approach

python/pyproject.toml
elchupanebrej: Exrta dependencies were declined

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,87 @@
...
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 20
```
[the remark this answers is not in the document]
elchupanebrej: Please leave pipeline timebound

codegen/templates/python.py.erb
```diff
@@ -0,0 +1,68 @@
+# This code was generated using the code generator from cucumber-messages.
```
luke-hill, line 1: Whilst it's not mandatory usually we split the two templates up into an enum template and a non-enum (Standard class), template.

Usually this is because the enums generate just some simple readers for each constant whereas the class usually has some primitive behaviour (like serialization / deserialization).

An example of the complexity difference in ruby would be these two

ENUM: Just has constant definitions https://github.com/cucumber/messages/blob/main/ruby/lib/cucumber/messages/attachment_content_encoding.rb

Class: Has some basic readers/serialization logic (Note the attr_readers and the .from_h method https://github.com/cucumber/messages/blob/main/ruby/lib/cucumber/messages/attachment.rb

luke-hill (commented):
I've reviewed 2/3rds of this PR so far and only got a couple of questions. So I'll pass that in for now. The other items are a bit more complex and I'll review at a later date (Might be after xmas now).

.github/workflows/test-codegen.yml
```diff
@@ -31,6 +31,11 @@ jobs:
         with:
           ruby-version: '3.1'
 
+      - name: Set up Python
```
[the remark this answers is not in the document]
elchupanebrej: No

.pre-commit-config.yaml
```diff
@@ -0,0 +1,50 @@
+# See https://pre-commit.com for more information
```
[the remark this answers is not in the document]
elchupanebrej: i could gitignore it for a while

codegen/templates/python.py.erb
elchupanebrej: The template is really simple, and all serialization/deserialization is done externally

python/pyproject.toml
```diff
@@ -0,0 +1,136 @@
...
+  "tox>=4.2"
+]
+test-coverage = [
+  "coverage",
+  "GitPython",
+  "packaging",
+  "pytest"
+]
+
+[project.scripts]
```
kieran-ryan, line 55: `project.entry_points` or `project.scripts` needed or intended to be populated?

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,63 @@
...
+          - os: macos-latest
+            python-version: "3.13"
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Generate code
```
luke-hill, line 46: This shouldn't be needed as this should be part of the regular code generation and checkin process

luke-hill (changes requested):
Reviewed and signed off 17/21 files, still got a few things to query

python/src/cucumber_messages/_messages.py
```diff
@@ -0,0 +1,954 @@
...
+     [IANA Media Type](https://www.iana.org/assignments/media-types/media-types.xhtml)
+     as well as Cucumber-specific media types such as `text/x.cucumber.gherkin+plain`
+     and `text/x.cucumber.stacktrace+plain`
+    """
+    media_type: str
+    """
+    *
+     Suggested file name of the attachment. (Provided by the user as an argument to `attach`)
+    """
+    file_name: Optional[str] = None
```
youtux, line 107: we can use `str | None` (we don't need to support python 3.8, which is EOL now)

python/src/cucumber_messages/_messages.py
```diff
@@ -0,0 +1,954 @@
...
+    source: Optional["Source"] = None
```
youtux, line 108: [nit] we can use `Source` instead of `"Source"`, since we use `from __future__ import annotations`

python/src/cucumber_messages/_messages.py
```diff
@@ -0,0 +1,954 @@
...
+
+
+@dataclass
+class Background:
+    description: str
+    id: str
+    keyword: str
+    """
+    The location of the `Background` keyword
+    """
```
youtux, line 227: [nit] can we inline this (and all the other similar ones?

python/src/cucumber_messages/_messages.py
```diff
@@ -0,0 +1,954 @@
...
+
+@dataclass
+class Comment:
+    """
+    *
+     A comment in a Gherkin document
+    """
+    """
+    The location of the comment
+    """
```
youtux, line 241: this will cause `Comment.__doc__`  to include the text "The location of the comment". We should better use comments (e.g. `# The location of the comment`)instead of strings to document attributes

python/src/cucumber_messages/_messages.py
youtux: actually, we should use comments, see https://github.com/cucumber/messages/pull/165/files#r1912240401

youtux:
I must have missed something. Why are we not using the code generator from https://github.com/koxudaxi/datamodel-code-generator anymore?

It would look more maintainable to me to use that one, as supporting the current proposed solution requires people that both both Ruby and Python, rather than just Python.

python/src/cucumber_messages/_messages.py
youtux: actually, I was wrong on this, it seems to work fine:
```
>>> class A:
...     """docstring of A"""
...     """docstring of foo"""
...     foo: str
...
>>> A.__doc__
'docstring of A'
```
Although it looks like the correct way is to put the docstring **after** the attribute declaration: https://stackoverflow.com/a/69754746/714760

.github/workflows/test-python.yml
```diff
@@ -0,0 +1,63 @@
...
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  build:
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
```
luke-hill, line 19:
```suggestion
      matrix:
        include:
          # Test latest python on windows / macos
          - { os: 'windows-latest', python-version: '3.13' }
          - { os: 'macos-latest', python-version: '3.13' }
        os: ['ubuntu-latest]
        python-version: ['3.9',, '3.10', '3.11', '3.12', '3.13', 'pypy3.9', 'pypy3.10']
```

codegen/generators/python.rb
```diff
@@ -0,0 +1,159 @@
...
+
+      if type.start_with?('list[')
+        list_type = type.match(/list\[(.*?)\]/)
+        inner_type = list_type[1]
+        if inner_type =~ /^[A-Z]/
+          "list[\"#{class_name(inner_type)}\"]"
+        else
+          "list[#{inner_type}]"
+        end
+      elsif type =~ /^[A-Z]/
```
luke-hill, line 42:
```suggestion
elsif type.match?(/\A[A-Z]/)
```

codegen/generators/python.rb
```diff
@@ -0,0 +1,159 @@
...
+      else
+        "#{snake_name}: Optional[#{property_type}] = None"
+      end
+    end
+
+    def get_property_type(parent_type_name, property_name, property)
+      type = type_for(parent_type_name, property_name, property)
+
+      if type.start_with?('list[')
+        list_type = type.match(/list\[(.*?)\]/)
```
luke-hill, line 35: Whilst this isn't wrong / buggy. Is there anything slightly more nuanced we can use for the inner capture. Maybe `(\w+)` ?

luke-hill: Also you can optimise this as you're only using the regexp matcher briefly

```
type[/list\[(.*?)\]/]
```

This will not store it in a variable, but it will store it in Regexp.last_match -> The index you want is `1` so then in the statements below you can call `Regexp.last_match(1)`

codegen/generators/python.rb
```diff
@@ -0,0 +1,159 @@
...
+        inner_type = list_type[1]
+        if inner_type =~ /^[A-Z]/
```
luke-hill, line 37: use .match? here too.

python/src/cucumber_messages/_messages.py
elchupanebrej: It's a 3.10 feature: https://peps.python.org/pep-0604/

.github/workflows/test-python.yml
elchupanebrej: It's needed for test execution later: We have to know if the model is fine for serialization/deserialization

luke-hill: Sorry I think you're misunderstanding. The code "itself" is checked in (See this PR below for the code `_messages.py` I think is the single file with all messages.

As part of your regular development flow you can develop new models and/or new facilities for enum's e.t.c. - Then once you run `make generate` it will generate all your new code. This is the same flow as every other language

This code (Running `make generate`), would therefore "re"generate the code on each and every PR which isn't something we would want to do, as it would create a rather large overhead. So you should just run `make generate` whenever you want to update your messages

codegen/generators/python.rb
youtux: Isn't there a better way rather than doing regexp match? I don't know ruby idioms, but in python I would return a boxed object that wraps the subtype, like `ArrayMarker(foo)`, and do an isinstwnce check here.

luke-hill: I am not sure what you are trying to do here as it looks quite complicated.

Basically storing any regex details inside a `#match` object is the heaviest footprint. And storing things briefly using Regexp.last_match is a "tiny" bit lighter, and Ideally just using `#match?` is the best because this doesn't store anything

luke-hill: If it's easier and you want me to tackle this as a follow-up can you make an issue and link to this given the size of this PR is quite large and first iteration is always the hardest one.

python/src/cucumber_messages/_messages.py
elchupanebrej: Done

.github/workflows/test-python.yml
elchupanebrej: Fixed

python/pyproject.toml
elchupanebrej: Fixed

codegen/templates/python.py.erb
elchupanebrej: Fixed

python/pyproject.toml
```diff
@@ -0,0 +1,132 @@
...
+test-coverage = [
+  "coverage",
+  "GitPython",
+  "packaging",
+  "pytest"
+]
+
+[tool.black]
+# Don't include autogenerated file
+force-exclude = ".*\\/src\\/cucumber_messages\\/_messages\\.py"
```
luke-hill, line 55: Is this line required now? This file **is** being checked in

python/pyproject.toml
```diff
@@ -0,0 +1,132 @@
...
+show_error_codes = true
+warn_return_any = true
+warn_unused_configs = true
+
+[[tool.mypy.overrides]]
+ignore_missing_imports = true
+module = [
+]
+
+# Wait until is ready: https://github.com/tox-dev/tox/issues/999
```
luke-hill, line 79: From a non-maintainers POV I can't tell if this is a bug or whatnot? The ticket in question looks like a new feature? Is this something we have / can have?

python/pyproject.toml
```diff
@@ -0,0 +1,132 @@
...
+[tool.tox]
+# language=INI
```
luke-hill, line 81: Is this meant to be commented out? or can it be removed?

luke-hill (commented):
22/23 files I'm good with. Few questions raised on final one - but might all be redundant. Top work.

As an aside - as/when I come to refactor the templates/generators, do you want the messages in 1 file per class like they are in some of the other languages. Not important for now, but just thought I'd make you aware that's quite easy with a bit of chopping that we do (Not required for this PR)

python/pyproject.toml
elchupanebrej: Yes, it's needed for other not generated python files

python/pyproject.toml
elchupanebrej: It's a helper comment for some IDEs to highlight string internal syntax

python/pyproject.toml
luke-hill: How is this?

```suggestion
# Once https://github.com/tox-dev/tox/issues/999 is released and available, migrate to the new tox approach
```

youtux:
> > I must have missed something. Why are we not using the code generator from https://github.com/koxudaxi/datamodel-code-generator anymore?
> 
> > 
> 
> > It would look more maintainable to me to use that one, as supporting the current proposed solution requires people that both Ruby and Python, rather than just Python.
> 
> 
> 
> - Python environment to generate code was unwanted
> 
> - There was an extra dependency

The extra dependency would only be a build-time dependency, so it wouldn't be needed by downstream users.

luke-hill (approved):
Top stuff. Please get a review from Rien/David also as they are the primary maintainers for Java/JS.

I don't foresee any major issues with other flavours such as go/dotnet e.t.c.

PR #216, opened 2024-04-23
[Ruby] Gem Bump

luke-hill:
#### Progressively updating lots of the supported packages.

- cck has gone from v12 to v15
- Rubocop subgems have all been updated 6-9months newer (Still a WIP)
- Rubocop Performance introduced to codebase (Currently not actively remedied)
- Rake / RSpec updated to latest support
- Ruby minimum gone from `2.5` to `3.0`
  - This has introduced new cop offenses and the ability to use Ruby 2.6+ and Ruby 3.0+ idioms
- Minimum Rubygems updated to 3.2.3 to solve CI niggles

#### Updated file formats for non-generated files to be 1 class per file
- ID Generation split into 2 files (UUID/Incrementing)

#### Message class simplified
- Deserialization logic was very simple and only used in the Message class so this has been absorbed and helper module removed
- Utils logic was only included in 2 areas (Only used in 1), so this also has been removed as a helper module and absorbed into the class proper

#### Miscellaneous changes
- Acceptance tests now recurse correctly through all CCK examples instead of having a procedural code execution for the whole glob of CCK examples
- The dummy messages used in a few specs have now been correctly setup inside `spec/support` and are now in a place to be properly extracted further into shared context's when the time allows

Actually keeping things up to date

### 📋 Checklist:

----


ruby/cucumber-messages.gemspec
```diff
@@ -10,8 +10,8 @@ Gem::Specification.new do |s|
   s.homepage    = "https://github.com/cucumber/messages#readme"
   s.platform    = Gem::Platform::RUBY
   s.license     = "MIT"
-  s.required_ruby_version = ">= 2.5"
-  s.required_rubygems_version = ">= 3.0.0"
+  s.required_ruby_version = ">= 2.7"
```
luke-hill, line 13: This will be set to 3.0 at the point this is ready to ship. This is just interim work to get it half working

Issue #72, opened 2020-10-18
ndjson message - Add hook type to message


grasshopper7:
**Is your feature request related to a problem? Please describe.**
The hook type (before, beforestep .... etc) is not available in the ndjson message. Need it in the reports to identify the hook type. Also for enabling or disabling display based on hook type.

**Describe the solution you'd like**
Available in the hook message. Will make porting reports based on json easier, in which this information is available.

**Describe alternatives you've considered**

**Additional context**

luke-hill:
@mpkorstanje - I think you were working on something that would effectively close this as completed?

mpkorstanje:
I was. I needed these to implement a message protocol based json formatter. But I'm blocked on until I solve other problems. If any one wants this, feel free to pick up #224.

PR #224, opened 2024-05-24
Add hook type

mpkorstanje:
Hooks in Cucumber can either execute before/after a scenario or before/after a step. Depending on which implementation is used these are put into the json report as either all part of the `element.steps` field[1] or the `element.before` and `element.steps` fields[2][3].

So to create a json reports from messages the type of hook must be included in the message.

 1. https://github.com/cucumber/cucumber-json-converter/blob/main/testdata/cucumber-js/7.3.2/json/failing-before-hook.json
 2. https://github.com/cucumber/cucumber-json-converter/blob/main/testdata/cucumber-jvm/7.2.3/json/failing-before-hook.json
 3. https://github.com/cucumber/cucumber-json-converter/blob/main/testdata/cucumber-jvm/7.2.3/json/failing-after-step-hook.json

Fixes: #72

Feel free.

### 📋 Checklist:


mpkorstanje:
It will be a while. I'm doing this for the JSON formatter, but that thing is overly complicated. And I'm getting side tracked with other projects in the same vein that are easier.

luke-hill:
@mpkorstanje is this now redundant by https://github.com/cucumber/messages/pull/102 ?

Issue #230, opened 2024-06-26
dotnet build does not match schema for Attachment


clrudolphi:
### 👓 What did you see?

The jsonschema and the Messages.md page both describe Attachment as having a "Source" property of type 'Source'.
The dotnet build however has the "Source" property of type 'SourceReference'.

Assuming the schema is correct, the problem is on line 362 of the messages.proto file within the dotnet folder.
```
message Attachment {
  SourceReference source = 1;
  string testStepId = 2;
  string test_case_started_id = 3;
``` 

Is the schema correct? Or is the messsages.proto file correct? I note the same discrepancy in the elixer folder's version of the messages.proto file.

### ✅ What did you expect to see?

One would expect the implementation to match the schema definition.

mpkorstanje:
Due to practical problems with protobuf and the fact that in practice only json was used, we switched to generating dtos based on the json schema. These dtos are then serialised to json using what ever each language has available.

Unfortunately Cucumber is maintained mostly by volunteers (see also https://mattwynne.net/new-beginning) and as a group we currently lack the expertise, time and/or interest to keep dot net actively up to date.

However you would be welcome to help out. Given the similarities between dot net and Java I think this would be an good starting point to understand the process.

https://github.com/cucumber/messages/blob/2b20a380c2bc269f0f6573228ebd1b22ee0a913e/java/Makefile#L18-L25

In essence, a template is applied to the json schema. This generates a single file with class names. This file is then cut into multiple files which are the message class names.

We also have a weekly zoom meeting. We sometimes use this to pair program. So we could help you get started. Meeting details can be found here:

https://github.com/cucumber/community-calendar?tab=readme-ov-file

Come say hi?

clrudolphi:
Ok, if I'm understanding correctly, then the dotnet generated code is still using the old code gen approach and is not currently compliant with the Messages schema. Is that the current state? 

I've glanced at the Ruby scripts for code gen. I can add in the template for dotnet.

I've also read that there is a docker image for building? Where is the build/gen process documented? (I'm not a ruby dev and have no interest in installing the tools on my machine but more than willing to dev within a container) I'll need to know how to map the repo to a volume for the image or how to build the image with my additions and changes.

mpkorstanje:
I think you have an accurate understanding of the current situation.

When it comes to building the project, I would suggest looking at the `.github` folder for the exact details. But in principle the setup is quite uncomplicated.

[...]

PR #233, opened 2024-06-28
Refactor dotnet code gen

clrudolphi:
### 🤔 What's changed?

Rewrote the Dotnet code generation script to adopt the ruby erb gem approach used for other languages.
Modified the target .NET Framework settings by eliminating the use of .NET45 - targeting .NETStandard2.0 only.

Used a DevContainer so you'll see a devcontainer.json in the file set.

To bring the .NET implementation to parity with the java version.

Per the question I posted on issue #230, does the ordering of properties of a entity matter when serialized? The serialization I have does not match the sequence of properties shown in the CCK examples.

I've used a Microsoft provided devcontainer image. Do you mind? Or should that be changed to the github image?
### 📋 Checklist:

  - [X ] I have added/updated tests to cover my changes.
- [X ] Users should know about my change
  - [X ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----


dotnet/Makefile
```diff
@@ -1,18 +1,24 @@
...
-	mkdir -p ~/.nuget/packages/grpc.tools/1.20.1/tools/linux_x64
-	cp /usr/bin/protoc ~/.nuget/packages/grpc.tools/1.20.1/tools/linux_x64/protoc
-endif
-	touch $@
+help: ## Show this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make <target>\n\nWhere <target> is one of:\n"} /^[$$()% a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-clean:
-	rm -f Cucumber.Messages/Messages.cs .fixprotoc
+generate: require .generate-messages ## Generate c# code based on the schemas found in ../jsonschema and using the scripts in ../jsonschema/scripts for the generation
```
mpkorstanje, line 8: Now that dotnet is supported, you can also add it to the supported languages in `Makefile` in the root of the project.

dotnet/Cucumber.Messages.Specs/MessagesSpec.cs
```diff
@@ -1,31 +1,68 @@
...
+            var soureReference = new SourceReference("samples/minimal/minimal.feature.ts",
+                null, null, new Location(3, null));
+            var stepDefPattern = new StepDefinitionPattern("I have {int} cukes in my belly", StepDefinitionPatternType.CUCUMBER_EXPRESSION);
+            var stepDefinition = new StepDefinition("0", stepDefPattern, soureReference);
+            var env = new Envelope(null, null, null, null, null, null, null, null, stepDefinition, null, null, null, null, null, null, null, null);
+
+            var jsonOptions = new JsonSerializerOptions();
+            jsonOptions.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
+            jsonOptions.Converters.Add(new CustomEnumConverter<StepDefinitionPatternType>());
+            jsonOptions.DefaultIgnoreCondition = System.Text.Json.Serialization.JsonIgnoreCondition.WhenWritingNull;
```
mpkorstanje, line 29: These json options are non trivial and will have to be replicated by anyone who wants to serialize messages. So I think it would be prudent to provide a serializer and deserializer class as part of the messages library. This should hopefully prevent any mistakes from taking hold.

You may also want to look at some of the other utility classes that come with messages. 

https://github.com/cucumber/messages/tree/main/java/src/main/java/io/cucumber/messages

When comparing do note that unlike many other languages, Java does not have an out of the box json serializer  On the other hand, the libraries Java does have update quite frequently. To avoid this cascading down from messages, the `MessageToNdjsonWriter` is constructed with a `Serializer` interface instead. This allows us to keep the serializer library at the edge of the dependency tree.

mpkorstanje (changes requested):
Looks good over all.

But I'm a bit puzzled as to why the CI hasn't been activated. I would have expected some build failures. I'll look into that.

For the mean time, I left a few comments. The first two are for myself. The third one you may want to address.

mpkorstanje:
> Per the question I posted on issue https://github.com/cucumber/messages/issues/230, does the ordering of properties of a entity matter when serialized? The serialization I have does not match the sequence of properties shown in the CCK examples.

Answered: https://github.com/cucumber/messages/issues/230#issuecomment-2197433270. The order does not matter.

> I've used a Microsoft provided devcontainer image. Do you mind? Or should that be changed to the github image?

As long as that choice falls within the common practices in the .Net ecosystem that is okay. The goal is to make sure the next contributor familiar with .Net understands what is going on.

I also don't think the Github provided containers are available on Dockerhub.

dotnet/Cucumber.Messages.Specs/MessagesSpec.cs
clrudolphi: Thanks for the quick feedback. I'll take a look at whether the utility classes make sense in the dotnet ecosystem. 
I'll double check with Gaspar about the .net version (5 v 8). 
And I'll port over the Java test suite (or something similar).

clrudolphi:
Updated the PR to include a ChangeLog entry.
Revised the code gen script to house all the generated C# classes in a /generated folder.
Added support classes in alignment with what the Java implementation has (for serialization support and TimeStamp conversions).
Added tests for the support classes (again, similar in nature to what the Java implementation does)
Dropped target framework support for .NET Framework 4.x (leaving .NET Standard 2.0 as the target).

dotnet/Cucumber.Messages.Specs/MessagesSpec.cs
clrudolphi: Completed. Please take a look at what was added to the support classes to make serialization easier. It aligns with how the Java implementation works.

clrudolphi:
@mpkorstanje please add @gasparnagy as a reviewer of this PR. Thanks

.github/workflows/test-dotnet.yml
```diff
@@ -22,5 +22,6 @@ jobs:
 
       - name: run tests
         run: |
-          make
+         echo "No tests yet"
+         exit 1
```
mpkorstanje, line 26: Please add something sensible here.

dotnet/Cucumber.Messages.Specs/ProtocolVersionTest.cs
```diff
@@ -0,0 +1,19 @@
...
+using Xunit;
+
+namespace Cucumber.Messages.Specs
+{
+    public class ProtocolVersionTest
+    {
+        [Fact]
+        public void ProtocolVersionIsCorrect()
+        {
+            Assert.Matches(@"\d+\.\d+\.\d+(-SNAPSHOT)?", Cucumber.Messages.ProtocolVersion.Version);
```
mpkorstanje, line 16: Does .Net use snapshot versions?

clrudolphi: It is not a community standard, no. Version numbering conventions are team-specific. Most teams do use some kind of suffix to designate pre release builds. 

I blindly copied what the Java implementation was doing. Sorry. I should have investigated what the prior .Net implementation had done and carried that forward. 

I'll revise as needed.

.github/workflows/test-dotnet.yml
clrudolphi: i've modified that to invoke dotnet test

dotnet/Cucumber.Messages.Specs/ProtocolVersionTest.cs
clrudolphi: The prior version also used a suffix called SNAPSHOT; or at least that's what I can make of it (sorry, not well versed with github workflows). Open to changing that to whatever makes sense. I will ask Gaspar about it on Monday.

mpkorstanje: No worries. I think that's a bit of copy paste from Gherkin. 

You can leave the `.csproj` file as is for now. I'll make a separate item to clean it up. It also means we need to change our release scripts that match against `VersionNumber` instead of `Version`.

https://github.com/cucumber/polyglot-release/blob/75e8af2c3398433d7a01862a863d038aa7d22dcb/polyglot-release#L70

dotnet/Cucumber.Messages.Specs/Cucumber.Messages.Specs.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
```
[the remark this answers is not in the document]
gasparnagy: This is just our internal test project, not a deliverable. So this will not cause any issues. (But the other changes will be breaking anyway for .NET, I guess).

jsonschema/scripts/codegen.rb
```diff
@@ -214,6 +214,33 @@ def format_description(raw_description, indent_string: "")
   end
 end
 
+class Dotnet < Codegen
+  def initialize(paths)
+    language_type_by_schema_type = {
+      'integer' => 'long',
```
gasparnagy, line 220: Are we sure that the 'integer' has to be a long in our case? Wouldn't `int` be enough?

@mpkorstanje Is there an explicit reason for requiring a 8 byte number for the messages? As far as I understand, we use numbers for line can column positions, where the maximum of the 4-byte int 2,147,483,647, should be more than sufficient. Even Visual Studio [treats line numbers as 4 byte `int`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.text.itextsnapshotline.linenumber?view=visualstudiosdk-2022#microsoft-visualstudio-text-itextsnapshotline-linenumber). Working with `long` is cumbersome in .NET because all other API uses `int` and converting them back and force is painful.

So my suggestion would be to use `int` here.

gasparnagy (changes requested):
Thx for the code, it looks good. 

I added my notes. Most of them minor, there are only a few tricky questions.

dotnet/Cucumber.Messages/generated/Attachment.cs
```diff
@@ -0,0 +1,157 @@
...
+            Object.Equals(TestStepId, that.TestStepId) &&         
+            Object.Equals(Url, that.Url);        
+    }
+
+    public override int GetHashCode() 
+    {
+        int hash = 17;
+        if (Body != null)
+          hash = hash * 31 + Body.GetHashCode();
+        if (ContentEncoding != null)
```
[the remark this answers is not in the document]
clrudolphi: @gasparnagy Any concerns with the rather rudimentary hash algorithm I'm using here?

dotnet/Cucumber.Messages.Specs/ProtocolVersionTest.cs
clrudolphi: Generalized that regex to look for an optional dot separated set of alphanumeric identifiers per the Semver standard as commonly used in .net apps.

dotnet/Cucumber.Messages/Cucumber.Messages.csproj
```diff
@@ -18,8 +19,8 @@
     <Product>Cucumber.Messages</Product>
     <PackageId>Cucumber.Messages</PackageId>
     <Authors>Cucumber Ltd, TechTalk</Authors>
-    <Copyright>Copyright &#xA9; Cucumber Ltd, TechTalk</Copyright>
-    <Description>Protocol Buffer messages for Cucumber's and SpecFlow's inter-process communication</Description>
+    <Copyright>Copyright © Cucumber Ltd, TechTalk</Copyright>
+    <Description>Cucumber messages for Cucumber's and SpecFlow's inter-process communication</Description>
```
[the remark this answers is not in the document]
clrudolphi: Adopted your wording.

dotnet/Cucumber.Messages/Cucumber.Messages.csproj
```diff
@@ -32,21 +33,11 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Google.Protobuf" Version="3.27.2"/>
-    <PackageReference Include="Grpc.Tools" Version="2.64.0">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
-
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3">
+     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3">
```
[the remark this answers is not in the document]
clrudolphi: Removed

dotnet/Cucumber.Messages/ProtocolVersion.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Cucumber.Messages
+{
+    public class ProtocolVersion
+    {
+        public static string Version
+        {
+            get
+            {
+                return Assembly.GetExecutingAssembly().GetName().Version.ToString();
```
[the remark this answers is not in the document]
clrudolphi: Adopted your suggested code into that function.

dotnet/Cucumber.Messages/generated/Attachment.cs
```diff
@@ -0,0 +1,157 @@
...
+ * [Source](#io.cucumber.messages.Source) file. It can be used for:
+ *
+ * * Syntax errors during parse time
+ * * Screenshots captured and attached during execution
+ * * Logs captured and attached during execution
+ *
+ * It is not to be used for runtime errors raised/thrown during execution. This
+ * is captured in `TestResult`.
+ */
+// Generated code
```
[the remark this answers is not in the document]
clrudolphi: done

jsonschema/scripts/codegen.rb
clrudolphi: The same script generates the TimeStamp class which uses two longs to represent the number of seconds that have passed since 1 Jan 1970 and the number of partial seconds expressed as nanoseconds.
Using a signed int would work, but only for the next 13.5 years before the seconds field would overflow.
Would you all prefer that the expressive range of implementations be consistent? 
Would you prefer that we use Int32 for Location fields but Int64 for timestamp? (would require some fiddling with the script but I think do-able).

luke-hill:
Dependent on the timeframes involved this will maybe require a large conflict resolution as I'm in the process of refactoring the codegen (Only a small refactor, but lots of moving of files around), here: https://github.com/cucumber/messages/pull/228

I'm currently at a bit of an impasse in terms of contributing for the coming weeks, so if this is merged in in the next 1-2 weeks you'll have no issues and I'll fix up conflicts my end. But just letting you know :+1:

dotnet/Cucumber.Messages/generated/Attachment.cs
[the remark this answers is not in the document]
gasparnagy: No. I'm fine with that.

dotnet/Cucumber.Messages/Cucumber.Messages.csproj
```diff
@@ -32,21 +33,11 @@
...
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-  </ItemGroup>
-
-  <ItemGroup>
-    <Compile Remove="Messages.cs"/>
-    <Protobuf Include="../messages.proto" OutputDir="$(MsBuildThisFileDirectory)" GrpcServices="None"/>
+     <PackageReference Include="System.Text.Json" Version="8.0.3" />
```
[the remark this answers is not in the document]
gasparnagy: I see. It seems that Reqnroll depends on Microsoft.Extensions.DependencyModel v6.0.0 anyways. And that version depends on System.text.json v6.0.0. So we can safely depend on that version without changing the overall dependency chain. Would system.text.json v6.0.0 be enough for our needs?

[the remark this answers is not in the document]
clrudolphi: Yes, that will work. Updated PR.

dotnet/Cucumber.Messages/NdjsonSerializer.cs
```diff
@@ -3,23 +3,30 @@
 
 namespace Cucumber.Messages
 {
-    public class NdjsonSerializer
+    public static class NdjsonSerializer
     {
+        private static object lockObject = new object();
```
gasparnagy, line 8: This solution is probably good enough, but for such thread-safe initializations, I generally prefer using the `Lazy<T>` class, because the lock object version needs to aquire a lock for every single property get, even after the initialization period.

clrudolphi: Ah. I had completely forgotten about the option to use Lazy. (I'm too old school for such new sparkly things) 
I'll change it and resubmit.

mpkorstanje:
@luke-hill cheers. 

@clrudolphi @gasparnagy take your time, I won't be looking at opensource this week until the end of the weekend.

dotnet/Cucumber.Messages/NdjsonSerializer.cs
clrudolphi: Done. Revised the PR to make this use Lazy<T>

jsonschema/scripts/codegen.rb
mpkorstanje: I reckon the correct way of solving this would be to add the [`minimum` and maximum` keywords](https://json-schema.org/understanding-json-schema/reference/numeric#range) to the line and column fields. Then the code generation can determine the most appropriate type based on that.

dotnet/Cucumber.Messages/CucumberMessagEnumConverter.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Cucumber.Messages
+{
+    public class CucumberMessagEnumConverter<T> : JsonConverter<T> where T : struct, Enum
```
mpkorstanje, line 10: Versioning is easier if a library only exposes the essentials. Consider making this and any other classes not explicitly intended to be part of the API package private (or the .Net) equivalent.

dotnet/Cucumber.Messages/NdjsonSerializer.cs
```diff
@@ -0,0 +1,53 @@
...
+        public static string Serialize<T>(T message)
+        {
+            return JsonSerializer.Serialize(message, JsonOptions);
+        }
+        public static Envelope Deserialize(string json)
+        {
+            return NdjsonSerializer.Deserialize<Envelope>(json);
+        }
+
+        public static T Deserialize<T>(string json)
```
mpkorstanje, line 46: I would only ever expect users of this API to read and write envelopes. So `Serialize<T>` and `Deserialize<T>` can probably be package private.

dotnet/Cucumber.Messages/CucumberMessagEnumConverter.cs
clrudolphi: Agreed. I'll make this class internal to the Cucumber.Messages assembly.

dotnet/Cucumber.Messages/NdjsonSerializer.cs
clrudolphi: I had left it public in order to test it (via elements more granular than Envelopes). Now that we're over that hurdle, I'll remove the single such test and make these private.

clrudolphi:
I've also added a line to MessagetoNdjsonReader and Writer classes to Close their internally held streamwriter upon Dispose().

jsonschema/scripts/codegen.rb
clrudolphi: @gasparnagy Is it worth waiting for this change to the schema to move through the ecosystem? I'll do what ever you two think best.

dotnet/.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+# Please see the documentation for more information:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+# https://containers.dev/guide/dependabot
+
+version: 2
+updates:
+ - package-ecosystem: "devcontainers"
+   directory: "/"
+   schedule:
+     interval: weekly
```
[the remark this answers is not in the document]
luke-hill: Is this still needed?

dotnet/Cucumber.Messages.Specs/BasicMessageSerializationTests.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public class BasicMessageSerializationTests
+    {
+
+        [Fact]
+        public void SerializesAnEnvelopeToNDJSONCorrectly()
+        {
+            var stepDefinitionNDJSON = @"{""stepDefinition"":{""id"":""0"",""pattern"":{""source"":""I have {int} cukes in my belly"",""type"":""CUCUMBER_EXPRESSION""},""sourceReference"":{""location"":{""line"":3},""uri"":""samples/minimal/minimal.feature.ts""}}}
+";          
+            var soureReference = new SourceReference("samples/minimal/minimal.feature.ts",
```
luke-hill, line 22: sourceReference*

dotnet/Cucumber.Messages.Specs/MessagesTest.cs
```diff
@@ -0,0 +1,27 @@
...
+        public void ThrowsWhenRequiredFieldsAreMissing()
+        {
+            Assert.Throws<ArgumentNullException>(() => new Background( null, null, null, null, null, null));
+        }
+
+        [Fact]
+        public void IsValidWhenNoRequiredFieldsAreMissing()
+        {
+            // This should succeed as no constructor arguments to an Envelope are required.
+            var env = new Envelope(null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null);
```
luke-hill, line 24: Should this have an assertion like no error is raised?

dotnet/Cucumber.Messages.Specs/MessagesTest.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Cucumber.Messages.Specs
+{
+    public class MessagesTest
+    {
+
+        [Fact]
+        public void ThrowsWhenRequiredFieldsAreMissing()
+        {
+            Assert.Throws<ArgumentNullException>(() => new Background( null, null, null, null, null, null));
```
luke-hill, line 17: Not sure if you care about linting but space before `(`

dotnet/Cucumber.Messages.Specs/NdjsonStreamSerializationTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            var json = Encoding.UTF8.GetString(memoryStream.ToArray());
+            Assert.Equal("{}"+Environment.NewLine, json);
+        }
+
+        [Fact]
+        public void IgnoresEmptyLines()
+        {
+            MemoryStream memoryStream = new MemoryStream(Encoding.UTF8.GetBytes("{}\n{}\n\n{}\n"));
+            var enumerator= new NdjsonMessageReader(memoryStream).GetEnumerator();
```
luke-hill, line 43: `enumerator = `

dotnet/Cucumber.Messages/Cucumber.Messages.csproj
```diff
@@ -18,9 +19,9 @@
     <Product>Cucumber.Messages</Product>
     <PackageId>Cucumber.Messages</PackageId>
     <Authors>Cucumber Ltd, TechTalk</Authors>
-    <Copyright>Copyright &#xA9; Cucumber Ltd, TechTalk</Copyright>
-    <Description>Protocol Buffer messages for Cucumber's and SpecFlow's inter-process communication</Description>
-    <PackageTags>cucumber-messages cucumber specflow gherkin</PackageTags>
+    <Copyright>Copyright © Cucumber Ltd, TechTalk</Copyright>
+    <Description>Cucumber messages to represent source data and execution results in Cucumber-family tools</Description>
```
luke-hill, line 23: > JSON schema-based messages for Cucumber's inter-process communication

dotnet/Cucumber.Messages/NdjsonSerializer.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+using System.Text.Json;
+
+namespace Cucumber.Messages
+{
+    public class NdjsonSerializer
+    {
+        private static Lazy<JsonSerializerOptions> _jsonOptions = new Lazy<JsonSerializerOptions>(() =>
+        {
+            var jso = new JsonSerializerOptions();
```
luke-hill, line 11: I'm guessing this one is meant to be jso, is there maybe a better name for this to avoid confusion over a misname? `options` perhaps?

jsonschema/scripts/codegen.rb
```diff
@@ -214,6 +214,33 @@ def format_description(raw_description, indent_string: "")
...
+    }
+
+    super(paths, language_type_by_schema_type)
+  end
+
+  def array_type_for(type_name)
+    "List<#{type_name}>"
+  end
+
+  def format_description(raw_description, indent_string: "")
```
luke-hill, line 232: For now I've just got a couple of fixes to suggest here as the entire format is changing. But if you change this to single quotes and the item suggested below then that should be fine for now. Until the stuff gets merged

jsonschema/scripts/codegen.rb
```diff
@@ -214,6 +214,33 @@ def format_description(raw_description, indent_string: "")
...
+    return '' if raw_description.nil?
+
+    raw_description
+      .split("\n")
+      .map { |line| line.strip() }
```
luke-hill, line 237: .strip does not require a `()` argument here. Nor in the one below

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+// Changes to this file may cause incorrect behavior and will be lost if 
+// the code is regenerated.
+// ------------------------------------------------------------------------------
+
+namespace Io.Cucumber.Messages.Types;
+
+/**
+ * Represents the <%= class_name(key) %> message in Cucumber's message protocol
+ * @see <a href="https://github.com/cucumber/messages" >Github - Cucumber - Messages</a>
+ <%- if not (schema['description'] || []).empty? -%>
```
luke-hill, line 17: `%- unless (schema['description'] || []).empty? -%>`

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+ *
+<%= format_description(schema['description'], indent_string: '') %>
+<%- end -%>
+ */
+
+public sealed class <%= class_name(key) %> 
+{
+  <%- schema['properties'].each do |property_name, property| 
+      required = (schema['required'] || []).index(property_name) 
+      isValueType = type_for(class_name(key), property_name, property) == "long" || type_for(class_name(key), property_name, property) == "bool"
```
luke-hill, line 27: this logic feels quite complex and can mess with the ordering of the lines. Is there any way we can push this into the class proper?

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+  -%>
+    <%- if not (property['description'] || []).empty? -%>
```
luke-hill, line 29: `<%- unless (property['description'] || []).empty? -%>`

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+    <%- end -%>
+    <%- if isValueType && !required -%>
+    public Nullable<<%= type_for(class_name(key), property_name, property) -%>> <%= capitalize(property_name) %> { get; private set; }
+    <%- else -%>
+    public <%= type_for(class_name(key), property_name, property) -%> <%= capitalize(property_name) %> { get; private set; }
+    <%- end -%>
+  <%- end -%>
+
+  <%- if (schema['required'] || []).empty? -%>
+  <%- schema['properties'].each_with_index do |(property_name, property), index| -%>
```
luke-hill, line 42: this can become `.each` and you can omit the index parameter in the pipes

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+
+    public static <%= class_name(key) %> Create(<%= type_for(class_name(key), property_name, property) -%> <%= property_name %>) 
+    {
+        return new <%= class_name(key) %>(
+        <%- schema['properties'].each_with_index do |(property_name_2, property_2), index_2| -%>
+          <%- if property_name_2 == property_name -%>
+            Require<<%= type_for(class_name(key), property_name, property) -%>>(<%= property_name %>, "<%= capitalize(property_name) %>", "<%= class_name(key) %>.<%= capitalize(property_name) %> cannot be null")<%= index_2 < schema['properties'].length-1 ? ',' : '' %>
```
luke-hill, line 49:
```suggestion
            Require<<%= type_for(class_name(key), property_name, property) -%>>(<%= property_name %>, "<%= capitalize(property_name) %>", "<%= class_name(key) %>.<%= capitalize(property_name) %> cannot be null")<%= index < schema['properties'].length - 1 ? ',' : '' %>
```

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+    <%- end -%>
+  <%- end -%>
+
+  <%- if (schema['required'] || []).empty? -%>
+  <%- schema['properties'].each_with_index do |(property_name, property), index| -%>
+
+    public static <%= class_name(key) %> Create(<%= type_for(class_name(key), property_name, property) -%> <%= property_name %>) 
+    {
+        return new <%= class_name(key) %>(
+        <%- schema['properties'].each_with_index do |(property_name_2, property_2), index_2| -%>
```
luke-hill, line 47:
```suggestion
        <%- schema['properties'].each_with_index do |(property_name_2, property_2), index| -%>
```

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+          <%- if property_name_2 == property_name -%>
+            Require<<%= type_for(class_name(key), property_name, property) -%>>(<%= property_name %>, "<%= capitalize(property_name) %>", "<%= class_name(key) %>.<%= capitalize(property_name) %> cannot be null")<%= index_2 < schema['properties'].length-1 ? ',' : '' %>
+          <%- else -%>
+            null<%= index_2 < schema['properties'].length-1 ? ',' : '' %>
```
luke-hill, line 51:
```suggestion
            null<%= index < schema['properties'].length - 1 ? ',' : '' %>
```

jsonschema/scripts/templates/dotnet.dotnet.erb
```diff
@@ -0,0 +1,150 @@
...
+          <%- end -%>
+        <%- end -%>
+        );
+    }
+  <%- end -%>
+  <%- end -%>
+
+    public <%= class_name(key) %>(
+    <%- schema['properties'].each_with_index do |(property_name, property), index| 
+        isValueType = type_for(class_name(key), property_name, property) == "long" || type_for(class_name(key), property_name, property) == "bool"
```
luke-hill, line 61: If this is pushed into your class you then get the win here

luke-hill (commented):
Popped some linting suggestions. I've tried to keep them brief / relevant. 

I think it's probably best we get this merged in first and then merge my codegen stuff in after as there's still a lot of stuff to tidy up but probably best I tackle it.

jsonschema/scripts/codegen.rb
gasparnagy: No. Let's just keep them as `long`. It does not worth the effort.

dotnet/Cucumber.Messages/CucumberMessagEnumConverter.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Reflection;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Cucumber.Messages
+{
+    internal class CucumberMessagEnumConverter<T> : JsonConverter<T> where T : struct, Enum
```
gasparnagy, line 10: Just found a missing "e". CucumberMessag**e**EnumConverter

dotnet/Cucumber.Messages.Specs/BasicMessageSerializationTests.cs
clrudolphi: Fixed misspelled variable name.

Re: the comment above about "Is this still needed?" - does that refer to the dependabot itself or the use of the devcontainer?

dotnet/Cucumber.Messages.Specs/MessagesTest.cs
clrudolphi: I was duplicating the approach taken by the Java implementation. It demonstrates that no arguments are required to the main constructor of an Envelope.
The alternative here, IMO, would be to remove the test entirely.

dotnet/Cucumber.Messages.Specs/MessagesTest.cs
clrudolphi: Corrected.

dotnet/Cucumber.Messages.Specs/NdjsonStreamSerializationTests.cs
clrudolphi: Corrected.

dotnet/Cucumber.Messages/Cucumber.Messages.csproj
clrudolphi: Adopted your wording.

dotnet/Cucumber.Messages/NdjsonSerializer.cs
clrudolphi: Adopted the use of the variable name 'options'.

dotnet/Cucumber.Messages/CucumberMessagEnumConverter.cs
clrudolphi: Corrected.

jsonschema/scripts/codegen.rb
clrudolphi: indent_string is now ''

jsonschema/scripts/codegen.rb
clrudolphi: adopted

jsonschema/scripts/templates/dotnet.dotnet.erb
clrudolphi: adopted

jsonschema/scripts/templates/dotnet.dotnet.erb
clrudolphi: This cannot be pushed into the .NET class. The intent is to identify those properties and fields that are optional value types. The declarations for these need to be different than a regular long or bool; they are declared as Nullable<T> (see lines 35/37 and 65/67 for a comparison). 
I don't understand your meaning of "can mess with the ordering of the lines" - what is the impact? Is there an example within the code to demonstrate?

jsonschema/scripts/templates/dotnet.dotnet.erb
clrudolphi: adopted

jsonschema/scripts/templates/dotnet.dotnet.erb
clrudolphi: adopted

jsonschema/scripts/templates/dotnet.dotnet.erb
clrudolphi: adopted

clrudolphi:
Thanks for your review comments so far. 
I think the only outstanding item is the comment from @luke-hill about the dotnet generation script and the complexity required when declaring Nullable<T> properties/fields.

clrudolphi:
@mpkorstanje @luke-hill Any chance to close this out today or tomorrow? I'm going on holiday until the 1st of August and would be good to finish this so that @luke-hill can proceed with his other refactoring.

dotnet/Cucumber.Messages.Specs/BasicMessageSerializationTests.cs
luke-hill: The use of dependabot. As I know we were discussing using renovate which is used most other places.

jsonschema/scripts/templates/dotnet.dotnet.erb
luke-hill: Sorry I meant more migrating this into the `Dotnet` ruby class that would determine this logic so we can make the overall erb template smaller. But I think this is probably best done after this has been merged and then I tag you for review. It will probably make more sense that way and you've done a lot of work here I don't want to delay / overcomplicate things.

jsonschema/scripts/templates/dotnet.dotnet.erb
luke-hill: Ignore this comment @clrudolphi based on the above comment. We will tackle this after the fact once this is approved and merged.

luke-hill:
@clrudolphi consider all my comments that have not been addressed to be "deferable" I don't want to hold anything up.

Just consider the actual dotnet stuff from Gaspar / Rien now. I can do some more fine tuning after your stuff is merged in a separate PR entirely to both this one and my companion one. Which would probably make things easier for you to review also.

dotnet/Cucumber.Messages.Specs/MessagesTest.cs
mpkorstanje: There is an `assertDoesNotThrow(() -> { some code })` in Java, but its kinda redundant in single line test case.

mpkorstanje (approved):
Looks good! 

@gasparnagy any remarks? If not, feel free to merge.

gasparnagy (approved):
Shit it 😀!

clrudolphi:
> Shit it 😀!

Funniest merge approval comment ever!

gasparnagy:
😀 Autocorrect... But you know what I meant.

mpkorstanje:
@gasparnagy have you made a release before?

gasparnagy:
No, I don't think so.

luke-hill:
Just posting here. Let's hold off releasing until I get my codegen stuff in also. As then it's a chunky enough release for a major.

mpkorstanje:
No problem.

PR #241, opened 2024-07-30
Refactored so that the dependency on System.Text.Json is removed from Cucumber.Messages deployable.

clrudolphi:
### 🤔 What's changed?
Moved the dependency to the Specs testing project only. The Cucumber.Messages project no longer has a dependency on System.Text.Json.

.NET implementations of Cucumber either take a dependency on an older version of the System.Text.Json library or want to avoid consuming libraries (such as this one) that hardwire such dependencies.

### 📋 Checklist:

< CHANGELOG already includes the refactoring of the DotNet implementation as an unreleased item. I felt an additional entry just for this was unwarranted.>
----

### DIFF
--- a/dotnet/Cucumber.Messages.Specs/CucumberMessagEnumConverter.cs
+++ b/dotnet/Cucumber.Messages.Specs/CucumberMessagEnumConverter.cs
@@ -5,12 +5,12 @@ using System.Reflection;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
-namespace Cucumber.Messages
+namespace Cucumber.Messages.Specs
 {
     internal class CucumberMessageEnumConverter<T> : JsonConverter<T> where T : struct, Enum
     {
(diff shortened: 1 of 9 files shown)


mpkorstanje:
Cheers! Looks good.

mpkorstanje:
@clrudolphi Do you have any breaking changes in mind? If not, we can make a release.

@luke-hill I don't think you have any breaking changes in the pipeline.

clrudolphi:
No. No breaking changes as none of this had been previously released.

mpkorstanje:
No I meant, on top of the current state.  I would rather not make a new major release and 5 minutes later yet another. :smile:

clrudolphi:
Oh. No. No further changes planned.

Chris
________________________________

mpkorstanje:
Okay. Luke would like to finish his refactoring and then we can release.

https://github.com/cucumber/messages/pull/233#issuecomment-2275680454

Apologies for the hold up. Normally we can release more quickly because the project is relatively quite.

PR #248, opened 2024-09-07
Remove dependency on go fmt for code gen

mpkorstanje:
### 🤔 What's changed?

Removed dependency on go fmt when generating code by generating
properly formatted code.

Code generation should run with as few dependencies as possible to
make it easier for  all developers to use (i.e. without every one having to
install every dependency under the sun).

@luke-hill is this idiomatic enough?

### 📋 Checklist:


luke-hill (commented):
Do you want me to do styling reviews here. Or just push this in and then tidy up with another piece of work? Happy with either.

mpkorstanje:
I was hoping you'd see something horrible. But looks like it isn't so bad. 😁

PR #253, opened 2024-09-15
[.NET] Added .NET 8, NRT and more

Romfos:
### 🤔 What's changed?

- Update target frameworks: .NET 8, .NET Standard 2.0
- Run tests for .NET 8 + .NET Framework
- Adopt C# 12 syntax
- Added editorconfig
- Remove non exited files from solution
- enable warnings as errors
- Update .NET sdk for github actions

Template generator changes:
- Enable NRT support
- Switch from classes to records. Equals, GetHashCode, ToString will be autogenerated
- Remove copy-paste helpers for every class like Require & RequireNonNull methods
- Correct implementation for value types

This PR is result of discussion here https://github.com/cucumber/gherkin/pull/281

### 📋 Checklist:

----


Romfos:
Important:
This PR is a "source breaking change".
Nullable annotations could lead to new warnings in code that depends on this library.
For example if external code will try to pass null in not expected places.

functionally library is the same

.github/workflows/release-nuget.yaml
```diff
@@ -12,10 +12,12 @@ jobs:
     environment: Release
     steps:
       - uses: actions/checkout@v4
+      
       - name: Setup .NET
         uses: actions/setup-dotnet@v4
         with:
-          dotnet-version: 6.0.x
+          dotnet-version: 8.0.x
```
clrudolphi, line 19: @gasparnagy - are you OK with changing the framework version(s) for cucumber/messages? I think we're the only  consumers at this point. 
If I understand correctly, and please confirm, our policy is that anything published as a nuget package (such as Cucumber.Messages) should target .NET Standard 2.0 (only) and that these can be built and tested on .net 8?

codegen/templates/dotnet.dotnet.erb
```diff
@@ -71,83 +67,14 @@ public sealed class <%= class_name(key) %>
...
-        RequireNonNull<<%= type_for(class_name(key), property_name, property) %>>(<%= property_name %>, "<%= capitalize(property_name) %>", "<%= class_name(key) %>.<%= capitalize(property_name) %> cannot be null");
-      <%- end -%>
-      <%- if property['items'] -%>
-        <%- if required -%>
-        this.<%= capitalize(property_name) %> = new <%= type_for(class_name(key), property_name, property) %>(<%= property_name %>);        
-        <%- else -%>
-        this.<%= capitalize(property_name) %> = <%= property_name %> == null ? null : new <%= type_for(class_name(key), property_name, property) %>(<%= property_name %>);
-        <%- end -%>
+      <%- if !isValueType && required -%>
+        <%= capitalize(property_name) %> = <%= property_name %> ?? throw new ArgumentNullException("<%= capitalize(property_name) %>", "<%= class_name(key) %>.<%= capitalize(property_name) %> cannot be null");
```
clrudolphi, line 73: The template was generating "record-like" classes with getters only; in a sense providing an 'immutable' data structure. This PR is adopting the use of records directly, making a stronger statement about immutability.
During the constructor, we previously copyied the contents of enumerable items by calling 'new List<T>(argument) for those properties that were enumerable. This disconnected the Message instance from its source (if one list or the other got changed, the other would not).

This proposed template simply assigns the input argument to the generated property.

@gasparnagy How careful do we need to be about immutability and/or effects of change here?
We could use an ImmutableList<T> here to be explicit about intent and strengthen the behavior, go back to previous behavior of simply copying the list content, or accept this behavior of accepting the List argument as-is.

.github/workflows/release-nuget.yaml
Romfos: this line change only sdk version for release github action.
.NET standard 2.0 is still in tfm list

codegen/templates/dotnet.dotnet.erb
Romfos: Immutable array is desirable here, but i will require 1 extra dependency for .NET Standard tfm

clrudolphi:
Yeah. But Gaspar and I had previous conversations about this library and at the time had agreed to limit ourselves to .net 6 as the baseline. Let's let him weigh in tomorrow.

Chris
________________________________

.github/workflows/release-nuget.yaml
gasparnagy: Yes, this is fine and as @Romfos mentioned this is just what SDK we use to build. As long as we target .NET Standard 2.0 it should be all good.

codegen/templates/dotnet.dotnet.erb
gasparnagy: I think regardless of what structure we use to store the "immutable" list, we should go back to the previous model of "copying" the items from the input structure, so the simple assignment is not enough. 

As this is package is not really used yet, maybe we should be more future proof and go for the ImmutableList accepting the fact that this way we have to add a dependency to `System.Collections.Immutable` for the .NET Standard 2.0 target (in .NET 8 it is anyway included). So I would say, let's give it a try. If we see problems with this dependendency, we can still roll back.

dotnet/Cucumber.Messages/Cucumber.Messages.csproj
```diff
@@ -1,7 +1,7 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
+	
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
-    <LangVersion>10.0</LangVersion>
+    <TargetFrameworks>net8.0;netstandard2.0</TargetFrameworks>
```
gasparnagy, line 4: I have gone through this a few times, but I still not fully sure what benefit do we have with explicitly including .NET 8.0 as a target framework? My view on this so far was that this would be only useful (in a library!) if we would need any `#if` pragma in the code for the different targets.
@Romfos Could you please detail your intention a bit more in detail about the explicitly listed target?

gasparnagy (changes requested):
I think the overall goal and the concept of the PR is fine.
See my comment on the explicit listing of the .NET 8 target and also we should clean up the "immutable list" problem mentioned by @clrudolphi (see discussion there).

codegen/templates/dotnet.dotnet.erb
Romfos: PR is updated

dotnet/Cucumber.Messages/Cucumber.Messages.csproj
Romfos: Yes, for example. We have method GetPackageVersion

image

We can make a build for .NET Standard 2.0 tfm and .NET 8 tfm

image
As you can see in one case we have an error, in second - we don't

Why this is working like that:
.NET 8 BCL and libraries have nullable reference type annotations. .NET Standard 2.0 is old and don't have it. By this reason .NET 8 version knows that GetCustomAttribute could return null, .NET Standard version - dont't know about it.
By this reason in .NET 8 version have additional warning about it

Why make sense to have a .NET 8 tfm in a list, event if you don't have a `#if` directives:
1) nullable reference types that lead to additional warnings (if Nullable enabled)
2) In general .NET team implemented bunch of new compiler warnings that help write more reliable code. Part of them could be backported to .NET Standard\.NET Framework. Some part - not
3) Compiler could generate more optimal IL code for modern .NET
4) New language features like `init` keyword, `requred` keyword, init records, e.t.c.

gasparnagy: Thx! I see, and let's keep it as it is then, but essentially some of these features are only about compiling our own code (so we could make a build for `-f net8.0` just to get the warnings, but do not include this into the package) or something (like `init`) that we anyway cannot use is we need to compile for .NET Standard as well. So the only one that is really applicable is the more optimal IL, but on the other side, the compilation time and the package size is doubled. Obviously for such small package all this does not really matter, it was more about my curiosity. Let's keep the double target. 😀

codegen/templates/dotnet.dotnet.erb
gasparnagy: Thx. As far as I see now it takes `ImmutableArray` as an input, so it will be the responsibility of the caller to ensure the immutability. I think this is good, but I wonder if we should generate an additional convenience overloads for the ctors that just take an `IEnumerable<T>` and build the `ImmutableArray` from it.
@clrudolphi Would creating an `ImmutableArray` for calling the ctors cause any issues for you?

CHANGELOG.md
```diff
@@ -8,8 +8,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+### Added
+- [Dotnet] Run tests for .NET 8 + .NET Framework 4.6.2
+- [Dotnet] Added editorconfig
+
 ### Changed
 - [Go] Switch to Google's UUID module ([#251](https://github.com/cucumber/messages/pull/251))
+- [Dotnet] Update target frameworks: .NET 8, .NET Standard 2.0
+- [Dotnet] Adopt C# 12 syntax
+- [Dotnet] Enable warnings as errors
+- [Dotnet] Enable nulalble for all projects
```
clrudolphi, line 21: spelling of nullable

codegen/templates/dotnet.dotnet.erb
clrudolphi: I would prefer the constructor signature to accept the IEnumerable<T> as a parameter, and have it create the ImutableArray property.

CHANGELOG.md
Romfos: good catch. fixed

codegen/templates/dotnet.dotnet.erb
Romfos: I disagree that we need to make a copy inside of constructor without any visible reason. 
Maybe you have some code examples and I can propose a better solution?

Romfos: okay, then lets continue to use current solution with List copy

mpkorstanje:
@Romfos did you close this by accident?

Romfos:
> @Romfos did you close this by accident?

no, accroding to comment https://github.com/cucumber/messages/pull/253#discussion_r1765224345 we will save current implementation

mpkorstanje:
Thanks for explaining!

clrudolphi:
@Romfos Oh, I had missed that this had gotten closed. Even if we leave the collection class type unchanged, I would like to see the other changes proposed within this PR adopted.

PR #301, opened 2025-06-07
Fix wrong reference for global attachments

davidjgoss:
### 🤔 What's changed?

I made a design mistake in https://github.com/cucumber/messages/pull/102.

As a result, the schema supports attachments that are at the test run level - because they are made from a `BeforeAll` or `AfterAll` hook - but links them directly to the test run via `testRunStartedId`, when really it should be to the hook with `testRunHookStartedId`, which in turn can be traced to the test run if required.

In practise, no consuming projects have implemented test run level attachments with these messages yet - either in formatters or Cucumber implementations. Nevertheless we'll do this in an expand-migrate-contract style, with the old deprecated field being removed later. This does still count as a breaking change, because for some languages any new field means a changed constructor signature.

### 📋 Checklist:

----

### DIFF
--- a/dotnet/Cucumber.Messages/generated/Attachment.cs
+++ b/dotnet/Cucumber.Messages/generated/Attachment.cs
@@ -58,11 +58,11 @@ public sealed class Attachment
     public string MediaType { get; private set; }
     public Source Source { get; private set; }
     /**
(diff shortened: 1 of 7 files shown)


mpkorstanje (commented):
Looks good! Just a doc comments.

mattwynne (approved):

PR #305, opened 2025-07-02
Add timestamp to Attachment

### LINKED ISSUE #304: Add Attachment.Timestamp to the schema
### 🤔 What's the problem you're trying to solve?

During test and step executions the execution might produce multiple attachments. For longer steps the exact timing of when the attachments were captured might be important.
Currently the `Attachment` envelop type does not have a timestamp and therefore this information cannot be captured and used.

### ✨ What's your proposed solution?

Add a `timestamp` field for the `Attachment` type just like in any other execution-related envelope types, like `TestStepStarted`.

### ⛏ Have you considered any alternatives or workarounds?

As the `Attachment` type does not have any extensibility point to capture additional information, I don't see any meaningful workaround. 
The timestamp can be set as optional for backwards compatibility.

### 📚 Any additional context?

Needed to properly capture attachments for Reqnroll.

//CC @clrudolphi

davidjgoss:
### 🤔 What's changed?

Add `Attachment.timestamp`

Fixes #304.

> During test and step executions the execution might produce multiple attachments. For longer steps the exact timing of when the attachments were captured might be important. Currently the Attachment envelop[e] type does not have a timestamp and therefore this information cannot be captured and used.

It makes sense to do this now because we already have the signature change for Attachment in #301 which isn't released yet.

### 📋 Checklist:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## [Unreleased]
 ### Added
 - Add `Attachment.testRunHookStartedId` for traceability of attachments to test run hooks ([#301](https://github.com/cucumber/messages/pull/301))
+- Add `Attachment.timestamp` ([#305](https://github.com/cucumber/messages/pull/305))
 
 ### Fixed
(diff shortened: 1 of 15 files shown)


mpkorstanje (approved):
LGTM, but this is a breaking change.

PR #306, opened 2025-07-07
Fix C++ version and ABI/SONAME version for 28.0.0

musicinmybrain:
### 🤔 What's changed?

Updated the C++ version numbers to match the overall release version number in https://github.com/cucumber/messages/releases/tag/v28.0.0.

All of the C++ versions in the 28.0.0 release are based on the previous 27.2.0 release, which is wrong, confusing, and misleading.

### 📋 Checklist:

It’s possible that this should be added to the changelog, but it depends on how you want to handle it.

----

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -6,7 +6,7 @@
 ###
 cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
-project(cucumber_messages VERSION 27.0.2 LANGUAGES C CXX)
+project(cucumber_messages VERSION 28.0.0 LANGUAGES C CXX)
 
 ###
 #
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -1,12 +1,22 @@
-###
-#
-# Project
-# name and version
-#
-###
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+# The VERSION here is a placeholder; the real version is in the VERSION file.
+set(VERSION 28.0.0)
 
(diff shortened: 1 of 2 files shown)


musicinmybrain:
> LGTM. Would it be possible to use the version from the VERSION file? It gets updated by the release process automatically.
> 
> https://github.com/cucumber/messages/blob/main/cpp/VERSION

We can certainly try something like https://github.com/cucumber/gherkin/commit/f8814c442771960f3659b619a48cf991dab79cb6. As discussed in https://github.com/cucumber/gherkin/pull/376#discussion_r1985065986, it has the “wart” that there is a hard-coded bogus placeholder `VERSION` that is replaced by the one in the `VERSION` file, which can be confusing, but it does seem to work.

musicinmybrain:
> The follow-up commit I just pushed should read the `VERSION` file correctly.

No, I was in too much of a hurry, and I didn’t deal with `cpp/src/lib/messages/CMakeLists.txt`.

musicinmybrain:
There, that should be right now!

mpkorstanje:
>  it has the “wart” that there is a hard-coded bogus placeholder VERSION that is replaced by the one in the VERSION file, which can be confusing, but it does seem to work.

As long as:

 a. the version can be reliably replaced with a `sed` command
 b. the version is located in "well known" files"
 c. the project files are standardized across all cpp implementations (gherkin and messages)
 
 Then we could remove that wart.
 
 The `VERSION` file is currently updated by  [polyglot-release](https://github.com/cucumber/polyglot-release)  and that script could take `cpp/CMakeLists.txt` into account as well.

PR #308, opened 2025-07-17
Add location to pickle

mpkorstanje:
### 🤔 What's changed?

Add (optional) `location` field to the `Pickle` message.

Finding the source location of a pickle without having to dive into the Gherkin AST and/or bring in one of the query libraries. Helps with, among other things:

- https://github.com/cucumber/query/pull/114
- https://github.com/cucumber/cucumber-node/pull/149

### 📋 Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,7 +26,7 @@ repos:
     hooks:
       - id: mypy
   - repo: https://github.com/astral-sh/uv-pre-commit
-    rev: 0.9.9
+    rev: 0.9.10
     hooks:
       # Ensure the uv lock file is up to date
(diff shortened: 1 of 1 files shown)


mpkorstanje:
> In all other objects that contain a Location, by convention, the Location is the first property mentioned and thus the first parameter in the generated constructors. This proposed change has the Location as the last property mentioned; and thus the last parameter in the parameter list. This inconsistency is a bit jarring. Can we move it up to the top of the property list?

Sure!

> For Examples(Scenario Outlines), what is the proper Location of the pickle resulting from the table?

The pickle location is the same as the location of the example row used to create the pickle.

PR #310, opened 2025-07-17
java: Use Integers in addition to Longs

mpkorstanje:

Currently, messages are a bit cumbersome to use in combination with Java. The generated code currently uses long values where Java would use integers. This leads to some silly code:

```java
lines.add((int) (long) location.getLine());
```

Unfortunately, long values are needed for the timestamp and duration. So simply replacing longs with integers wouldn't work.

By specifying some upper bounds the code generation can make a more informed choice about the size of number to use. The choice for 2^31 - 1 is somewhat arbitrary but matches the maximum length of strings in Java and .Net. In practice, I would expect Gherkin documents to have a much more reasonable length though.

Partially resolves: https://github.com/cucumber/cucumber-jvm/issues/3013

Does anyone have a better idea? 

### 📋 Checklist:
- [x] https://github.com/cucumber/messages/pull/311 to make this PR smaller.

### DIFF
--- a/java/src/test/java/io/cucumber/messages/JacksonTest.java
+++ b/java/src/test/java/io/cucumber/messages/JacksonTest.java
@@ -29,7 +29,7 @@ class JacksonTest {
 
     @Test
     void can_deserialize_envelope() throws JsonProcessingException {
-        Envelope source = Envelope.of(new TestRunStarted(new Timestamp(3L, 14L), UUID.randomUUID().toString()));
+        Envelope source = Envelope.of(new TestRunStarted(new Timestamp(3L, 14), UUID.randomUUID().toString()));
         String json = OBJECT_MAPPER.writeValueAsString(source);
         assertEquals(source, OBJECT_MAPPER.readValue(json, Envelope.class));
     }


jsonschema/src/Location.json
```diff
@@ -5,10 +5,14 @@
   "description": "Points to a line and a column in a text file",
   "properties": {
     "line": {
-      "type": "integer"
+      "type": "integer",
+      "minimum": 0,
```
luke-hill, line 9: Should this be 1 too?

codegen/generators/php.rb
```diff
@@ -53,11 +53,11 @@ def nullable?(property_name, schema)
     end
 
     def scalar?(property)
-      property.key?('type') && language_translations_for_data_types.key?(property['type'])
+      property.key?('type') && select_language_translations_for_data_types(property['type'], property)
```
luke-hill, line 56: This will return `false` or a non-boolean value which I haven't checked downstream if this matters, but it could

codegen/generators/java.rb
```diff
@@ -23,12 +23,20 @@ def format_description(raw_description, indent_string: '')
...
 
-    def language_translations_for_data_types
-      {
-        'integer' => 'Long',
-        'string' => 'String',
-        'boolean' => 'Boolean'
-      }
+    def select_language_translations_for_data_types(type, property)
+      if type == 'integer'
+        if property['maximum'] and property['maximum'] <= 2147483647
```
luke-hill, line 28:
```suggestion
        if property['maximum']&.between?(1, 2147483647)
```

codegen/generators/base.rb
```diff
@@ -105,25 +105,23 @@ def property_type_from_ref(ref)
...
-        language_translations_for_data_types.fetch(type)
-      end
+      property
     end
 
     def type_for(parent_type_name, property_name, property)
       if property['$ref']
         property_type_from_ref(property['$ref'])
       elsif property['type']
-        property_type_from_type(parent_type_name, property_name, property, type: property['type'])
```
luke-hill, line 126: I assume this is the only location for this? If so this seems good. I think a lot of the initial work I did for codegen was converting into a generic structure - I didn't go through and tidy up. So good catch if so

luke-hill (commented):
Reviewed about half of the file. I think some bits are ok. But other bits might cause issues where they change the response structures. But I'm not 100% sure without digging in again

PR #315, opened 2025-08-11
Update documentation for `TestStepResult.message`

davidjgoss:
### 🤔 What's changed?

Expand on the documentation for `TestStepResult.message` to be clear that it should include the exception if there was one.

See https://github.com/cucumber/pretty-formatter/issues/12

### 📋 Checklist:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+- Update documentation for `TestStepResult.message` ([#315](https://github.com/cucumber/messages/pull/315))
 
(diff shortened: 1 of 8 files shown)


clrudolphi:
I'm confused by this. Why suggest that the Message property contain the content of an Exception when this type already has an Exception property?

mpkorstanje:
The `TestStepFinished.message` came first. But it couldn't be used in tools that need the type and message separately from the stacktrace so the `TestStepFinished.exception` was added later in https://github.com/cucumber/messages/pull/122. 

If you're worried about duplicate data, the message field is optional so you can omit it if you have a stacktrace.

gasparnagy:
> @gasparnagy Should Reqnroll split the .net exception into info about the type/message and the stack trace and populate these two properties respectively?

Let's discuss this at https://github.com/cucumber/messages/issues/317

PR #329, opened 2025-09-04
Add Suggestion message

mpkorstanje:

Cucumber implementations helpfully generate code to implement undefined steps. As we move to message based formatters, they should be able to provide these snippets to the user as well.

I didn't quite follow the design in https://github.com/cucumber/common/issues/2272 and I've aggregated the snippets for each pickle step into a single suggestion. This makes it easier to use the snippet type internally as the snippet generator can now generate snippets without knowing about the other parts of the event.

This does come with an implicit assumption that there is one suggestion per execution per undefined step. Does that assumption hold?

### 📋 Checklist:


clrudolphi (approved):
LGTM

PR #330, opened 2025-09-05
Clarify documentation around stepMatchArgumentsLists

mpkorstanje:

This has confused the hell out of me for the longest time.

### 📋 Checklist:

### DIFF
--- a/cpp/include/messages/cucumber/messages/test_step.hpp
+++ b/cpp/include/messages/cucumber/messages/test_step.hpp
@@ -16,8 +16,11 @@ using json = nlohmann::json;
 // Represents the TestStep message in Cucumber's message protocol
 // @see <a href=https://github.com/cucumber/messages>Github - Cucumber - Messages</a>
 //
-// A `TestStep` is derived from either a `PickleStep`
-// combined with a `StepDefinition`, or from a `Hook`.
+// A `TestStep` is derived from either a `PickleStep` combined with a `StepDefinition`, or from a `Hook`.
+//
+// When derived from a PickleStep:
+// * For `UNDEFINED` steps `stepDefinitionIds` and `stepMatchArgumentsLists` will be empty.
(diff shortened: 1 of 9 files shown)


clrudolphi:
That is not at all what I understand.

For instance, in the CCK Minimal example, the TestCase message is:
```
{"testCase":{"id":"6","pickleId":"3","testSteps":[{"id":"7","pickleStepId":"2","stepDefinitionIds":["4"],"stepMatchArgumentsLists":[{"stepMatchArguments":[{"group":{"start":7,"value":"42","children":[]},"parameterTypeName":"int"}]}]}],"testRunStartedId":"5"}}
```

and in the Parameter-Types example it is:
```
{"testCase":{"id":"7","pickleId":"3","testSteps":[{"id":"8","pickleStepId":"2","stepDefinitionIds":["5"],"stepMatchArgumentsLists":[{"stepMatchArguments":[{"group":{"start":0,"value":"LHR-CDG","children":[{"start":0,"value":"LHR","children":[]},{"start":4,"value":"CDG","children":[]}]},"parameterTypeName":"flight"}]}]}],"testRunStartedId":"6"}}
```

The `StepMatchArgumentsLists` appear to be a list of items that describe how the arguments of the step align with the parameters of the StepDefinition method. In the example of the Parameter-Type, it provides an entry in the `StepMatchArguments` for each match group of the regex.

Unfortunately, the CCK has no scenarios that cover an ambiguous binding situation.

clrudolphi:
@mpkorstanje Thank you for the explanation. 

What threw me off was the comment added by revised schema that is thrown in to the generated code that implies that consumers may look at the count of entries in the `StepMatchArgumentLists` array to determine whether there is ambiguity or not.

There is similar wording already on the `stepDefinitionIds` array, which is sufficient IMO.

The comment to be placed on the `StepMatchArgumentLists` array might be your explanation above: 
```
For an ambiguous step, a testcase will have test step with multiple entries in stepDefinitionIds and stepMatchArgumentsLists. The first entry in the stepMatchArgumentsLists holds the list of arguments for the first matching step definition, the second entry for the second, etc.
```

Issue #285, opened 2025-02-06
Proposal to remove JavaScript classes and just generate interfaces


davidjgoss:
Currently the JavaScript code generated for messages represents each message object as a class, with implicitly public fields and an empty constructor. This requires two dependencies `reflect-metadata` and `class-transformer` to make deserialisation work with the `parseEnvelope` function.

The classes have no methods, and the only thing they do vs just having plain objects is to default missing non-optional properties to `''` (in case of strings) or new objects (for classes) which I think is a dubious practise anyway.

I'd propose we just generate TypeScript types/interfaces rather than classes, meaning the schema would carry no code overhead, and no dependencies, and (de)serialisation could just be handled with standard `JSON.parse` and `JSON.stringify` as JavaScript users are well accustomed to. As a data point, `cucumber-js` and our first-party formatters only work with messages as plain objects and never instantiates the classes.

The automatic empty string/object behaviour would go away - it's unclear if anyone is relying on this.

cc @vitalets @badeball as other consumers of this library - any thoughts?

vitalets:
I like it! In most cases I use `cucumber/messages` for types.
The only question is about enums - are you going to transform this to type as well?
```js
export declare enum TestStepResultStatus {
    UNKNOWN = "UNKNOWN",
    PASSED = "PASSED",
    SKIPPED = "SKIPPED",
    PENDING = "PENDING",
    UNDEFINED = "UNDEFINED",
    AMBIGUOUS = "AMBIGUOUS",
    FAILED = "FAILED"
}
```

badeball:
> which I think is a dubious practise anyway.

I agree and fully support this, as I doubt it will have any unintended consequences for my self. I'd much rather like to rely on the type system for ensuring non-null values (ie. using `??` operator) instead of built-in defaulting, but that's me.

The messages library tends to "linger" in the dependency tree (JS), with old versions and deprecation notices, causing some [noise](https://github.com/badeball/cypress-cucumber-preprocessor/issues/1215), so I'm happy to see it go away.

PR #431, opened 2026-05-22
Add Dart Implementation

Fried-man:
### 🤔 What's changed?

Adds an initial Dart implementation for Cucumber Messages

There is currently no officially maintained Dart implementation of Cucumber Messages. I'd like to use a Dart implementation to build out Gherkin dart support.

If the API surface needs updates to better conform with the other language implementations.

### 📋 Checklist:

----


codegen/generators/dart.rb
```diff
@@ -0,0 +1,85 @@
...
+  # Generator::Dart
+  # Automatic code generation overrides for the Dart programming language
+  class Dart < Base
+    FILE_HEADER = <<~HEADER.freeze
+      // This code was generated using the code generator from cucumber-messages.
+      // Manual changes will be lost if the code is regenerated.
+      // ignore_for_file: cast_nullable_to_non_nullable, public_member_api_docs
+      // ignore_for_file: sort_constructors_first
+      // ignore_for_file: unnecessary_null_in_if_null_operators
+    HEADER
```
mpkorstanje: For consistency, please move this to the template. That's where it is expected.

codegen/generators/dart.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    def class_name(ref)
+      base = File.basename(ref, '.schema.json')
+      return 'DurationMessage' if base == 'Duration'
+      return 'ExceptionMessage' if base == 'Exception'
```
mpkorstanje: If this is necessary, this could do with a comment to explain why these names are different. Though I assume Dart has some name spacing to prevent collisions in the first place, if so then using the message names would be preferable. People very familiar with messages, but not so familiar with Dart will occasionally have to make small changes to implementations that use Dart. This would likely trip them up.

dart/lib/src/generated/duration.g.dart
```diff
@@ -0,0 +1,28 @@
+// This code was generated using the code generator from cucumber-messages.
+// Manual changes will be lost if the code is regenerated.
+// ignore_for_file: cast_nullable_to_non_nullable, public_member_api_docs
+// ignore_for_file: sort_constructors_first
+// ignore_for_file: unnecessary_null_in_if_null_operators
+
+part of 'messages.dart';
+
+class DurationMessage {
+  final int seconds;
```
mpkorstanje, line 13: What is the range on integer values in dart?

dart/lib/src/id_generator.dart
```diff
@@ -0,0 +1,40 @@
...
+/// (`"0"`, `"1"`, …).
+///
+/// Intended for tests; not safe for production use across processes.
+IdGenerator incrementingIdGenerator() {
+  var next = 0;
+  return () => '${next++}';
+}
+
+/// Returns an [IdGenerator] that produces random RFC 4122 version 4 UUIDs.
+IdGenerator uuidV4IdGenerator() {
```
mpkorstanje: I don't think it is a good idea to implement and maintain our own UUIDv4 generator. I also don't think it is a good idea to add a dependency on library that does this.

Ideally the `IdGenerator` is an interface that users of this library implement. For example by delegating to an existing UUIDv4 generator implementation that they pull in as a dependency.

dart/lib/src/ndjson_message_stream.dart
```diff
@@ -0,0 +1,102 @@
...
+/// schema.
+///
+/// The optional [lineNumber] is used only to improve parse error context.
+///
+/// When [includeLineInErrors] is false, parse errors do not include input
+/// line content.
+Envelope parseEnvelopeJson(
+  String json, {
+  int? lineNumber,
+  bool includeLineInErrors = true,
```
mpkorstanje: Please default this to false. While not likely, it avoids accidental information disclosure type issues as well as logging of untrusted data.

dart/CHANGELOG.md
mpkorstanje, line 1: Please don't duplicate the CHANGELOG in the root of the project. This one will be overlooked if we make changes.

.github/workflows/test-dart.yml
```diff
@@ -0,0 +1,69 @@
...
+
+jobs:
+  test-dart:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+        dart: ["3.5.0", stable]
```
mpkorstanje: Please pin these to exact versions. This keeps CI stable.

dart/Makefile
```diff
@@ -0,0 +1,27 @@
...
+
+benchmark: ## Run Dart throughput benchmarks
+	dart run benchmark/ndjson_benchmark.dart
+
+lib/src/messages.dart: $(schemas) ../codegen/codegen.rb ../codegen/templates/dart.index.dart.erb
+	ruby ../codegen/codegen.rb Generator::Dart dart.index.dart.erb > $@
+
+.generate-messages: $(schemas) ../codegen/codegen.rb ../codegen/templates/dart.dart.erb
+	ruby ../codegen/codegen.rb Generator::Dart dart.dart.erb > Generated.dart.tmp
+	ruby -e "input = File.readlines('Generated.dart.tmp', chomp: true); current = nil; lines = []; flush = lambda { next unless current; File.write(File.join('lib/src', current), lines.join(\"\\n\") + \"\\n\"); lines = [] }; input.each do |line| if line.end_with?('.g.dart') && !line.include?(' '); flush.call; current = line; else; lines << line; end; end; flush.call"
```
mpkorstanje: I assume this splits the generated file?

If so, please use the csplit solution used by other implementations.

https://github.com/cucumber/messages/blob/521cb9de652e1961fbbd8a5faf9b1e026b8eebc5/java/Makefile#L21-L25

mpkorstanje (changes requested):
Heya, thanks for the PR! 

I've given this a quick look for big problems. I didn't spot any, but I'll have to give it a more detailed look later. 

Some general remarks:

1. I assume you're aware of https://github.com/cucumber/gherkin/issues/30 and https://github.com/cucumber/gherkin/issues/29.

2. We currently don't publish the Gherkin parser. I assume that is something you'd like to see happen in the future. That can be arranged though but might take some work.

3. I see that you've chosen to set default values for all required fields. Perhaps copied from the JavaScript implementation. We are trying to get rid of those for JavaScript (https://github.com/cucumber/messages/issues/285). It would make sense not copy that mistake into Dart. Ideally the constructor would throw an exception when trying to create message with missing required fields.

4. The generated messages sit in the same folder as the support code. This makes reviewing code quite annoying. And this may cause issues in the future if introduce a message that clashes with the support code. Can we separate these?


A few more detailed remarks below.

dart/Makefile
```diff
@@ -0,0 +1,27 @@
...
+.DEFAULT_GOAL = help
+
+help: ## Show this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make <target>\n\nWhere <target> is one of:\n"} /^[$$()% a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
+
+generate: require lib/src/messages.dart .generate-messages ## Generate Dart code from JSON schemas
+
+require: ## Check requirements for generation and checks
+	@ruby --version >/dev/null 2>&1 || (echo "ERROR: ruby is required."; exit 1)
+	@dart --version >/dev/null 2>&1 || (echo "ERROR: dart is required."; exit 1)
```
mpkorstanje: Dart should not be required to generate the code. I suspect that will go away once the bench mark tests are removed.

.github/workflows/test-dart.yml
```diff
@@ -0,0 +1,69 @@
...
+    defaults:
+      run:
+        working-directory: dart
+
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - uses: dart-lang/setup-dart@c7a31f6a04bb9bf94f538fc551f7ef9e40223317 # v1.7.1
```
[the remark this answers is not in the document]
mpkorstanje: Where did you get these hashes from? Have you used an LLM?

I also don't see any actions ran in https://github.com/Fried-man/messages/actions. Please this yourself before submitting.

codegen/generators/dart.rb
Fried-man: Moved to where this is expected

dart/lib/src/generated/duration.g.dart
Fried-man: From [first-party docs](https://api.flutter.dev/flutter/dart-core/int-class.html):

The default implementation of int is 64-bit two's complement integers with operations that wrap to that range on overflow.

**Note**: When compiling to JavaScript, integers are restricted to values that can be represented exactly by double-precision floating point values. The available integer values include all integers between -2^53 and 2^53, and some integers with larger magnitude. That includes some integers larger than 2^63. The behavior of the operators and methods in the [int](https://api.flutter.dev/flutter/dart-core/int-class.html) class therefore sometimes differs between the Dart VM and Dart code compiled to JavaScript. For example, the bitwise operators truncate their operands to 32-bit integers when compiled to JavaScript.

dart/CHANGELOG.md
Fried-man: Makes sense, removed

dart/Makefile
Fried-man: Updated to run using cpslit solution

dart/Makefile
Fried-man: Removed dart requirement. I was mainly using it to run `dart format` on the generated code. Not important to format generated code.

.github/workflows/test-dart.yml
Fried-man: Pinned to exact versions

dart/lib/src/id_generator.dart
Fried-man: Agree, updated to only be an interface.

codegen/generators/dart.rb
Fried-man: It collides with some foundational dart classes. I refactored to use the normal names & refactored code that depends on it with explicit references to the file the class was from. Ex:
```
import 'package:cucumber_messages/src/messages.dart' as messages;

/// Converts a Dart [Duration] to a Cucumber [messages.Duration].
///
/// The [duration] value is exported as whole seconds plus nanoseconds.
messages.Duration durationToDurationMessage(Duration duration) {
  final micros = duration.inMicroseconds;
  final seconds = micros ~/ Duration.microsecondsPerSecond;
  final nanos = micros.remainder(Duration.microsecondsPerSecond) * 1000;
  return messages.Duration(seconds: seconds, nanos: nanos);
}
```

dart/lib/src/generated/duration.g.dart
Fried-man: So basically:
| Platform | Underlying Representation | Minimum Value | Maximum Value |
|----------|---------------------------|---------------|---------------|
| Native   | 64-bit signed integer     | \(-2^{63}\)   | \(2^{63} - 1\) |
| Web      | 64-bit float (JS)         | \(-2^{53} + 1\) (precise) | \(2^{53} - 1\) (precise) |

dart/lib/src/ndjson_message_stream.dart
Fried-man: Updated default to false

dart/Makefile
Fried-man: and yeah it does split the generated files

Fried-man:
> Heya, thanks for the PR! 
> 
> I've given this a quick look for big problems. I didn't spot any, but I'll have to give it a more detailed look later. 
> 
> Some general remarks:
> 
> 1. I assume you're aware of https://github.com/cucumber/gherkin/issues/30 and https://github.com/cucumber/gherkin/issues/29.
> 
> 2. We currently don't publish the Gherkin parser. I assume that is something you'd like to see happen in the future. That can be arranged though but might take some work.
> 
> 3. I see that you've chosen to set default values for all required fields. Perhaps copied from the JavaScript implementation. We are trying to get rid of those for JavaScript (https://github.com/cucumber/messages/issues/285). It would make sense not copy that mistake into Dart. Ideally the constructor would throw an exception when trying to create message with missing required fields.
> 
> 4. The generated messages sit in the same folder as the support code. This makes reviewing code quite annoying. And this may cause issues in the future if introduce a message that clashes with the support code. Can we separate these?

Hey! Thanks for the review

1. I did see that. Planning on solving those.
2. I would like the gherkin parser to be published. Totally understand it may take some work & for the short-term I can have code that depends on it point to the repo itself rather than pub.dev
3. I was concerned that was going to be considered un-wanted behavior but thought it was fine since JS had it (Missed #285 ). I have updated it to require these values.
4. Sure, I moved the generated files (`*.g.dart`) to `dart\lib\src\generated`.

.github/workflows/test-dart.yml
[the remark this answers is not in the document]
Fried-man: Updated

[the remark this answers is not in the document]
Fried-man: Yeah I used an LLM and it hallucinated. Updated to a real hash.

I tried to run it manually from the actions tab on my fork but didn't see an option to. Realized this morning that I could just open a PR on my fork and run the CI there. [Here](https://github.com/Fried-man/messages/actions/runs/26287008778) is a passing run from that.

dart/CHANGELOG.md
Fried-man: Actually, re-added but had it point to the root changelog. [pub.dev](https://pub.dev/) (official dart package distributor) has [a scoring system](https://pub.dev/help/scoring) for packages and having a changelog is one of the criteria. While it is not required, it can make the package not appear well-maintained & there is a GUI for the changelog on the site for potential adopters to reference.

If you would like, I can remove it again but felt like having it serve to only reference the root changelog was a fine solution to prevent the issue of having to maintain multiple versions.

Fried-man: Also, looks like they require I include the current version number in the changelog or else there will be an error when validating publishing. I included that validation for publishing dry run in the CI to ensure it stays publishable but can understand it be annoying to have to bump the changelog value whenever the package version changes.

Fried-man:
Alright, I think I'm done lol. Latest dart run passing on my fork can be found [here](https://github.com/Fried-man/messages/actions/runs/26288940814/job/77383672765?pr=1).

mpkorstanje (commented):
Great! I don't have time for another review pass today.

But I was reading https://pubrelease.onepub.dev/creating-a-release and I see that the `pub_release` command is rather involved and makes a lot of assumptions about the work flow.

Because we're releasing 10+ different languages we don't exactly have a standard process. 

But it looks like this:

1. Prepare the release by updating the version numbers and making other changes on a dev-machine.
2. Tag these changes in Git.
3. Push the changes and tags to GitHub.
4. GitHub will trigger on the tags and start the release process for each action independently.

Can you tell me if the `pub_release` command is the only way to "publish your project to pub.dev"? We might have some experimenting to do to that get that work.

Fried-man:
> Can you tell me if the pub_release command is the only way to "publish your project to pub.dev"? We might have some experimenting to do to that get that work.

I believe the official publish path is [dart pub publish](https://dart.dev/tools/pub/cmd/pub-lish). I don't believe it automatically creates tags or bumps the version.

.github/workflows/test-dart.yml
```diff
@@ -0,0 +1,65 @@
...
+
+jobs:
+  test-dart:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+        dart: ["3.7.0", "3.12.0"]
```
mpkorstanje, line 32: What is the rationale for these two versions specifically?

dart/benchmark/ndjson_benchmark.dart
mpkorstanje, line 1: Please don't include a Benchmark. This is more code to maintain.
I also expect that the bulk of time is going to sit in jsonEncode/jsonDecode so I don't think this will be terribly informative.

dart/example/lib/main.dart
mpkorstanje, line 1: Please don't include examples. This is more code to maintain. The tests and readme should be sufficient examples.

dart/lib/src/id_generator.dart
```diff
@@ -0,0 +1,9 @@
+/// Generates a new unique identifier string.
+///
+/// Implement this in your application (for example by delegating to a
+/// UUID library you already depend on).
+// ignore: one_member_abstracts
+abstract interface class IdGenerator {
```
mpkorstanje: If I'm not mistaken, this class is now only uses in tests so it can be moved or removed.

dart/README.md
```diff
@@ -0,0 +1,154 @@
...
+final jsonLine = envelopeToJsonString(envelope);
+
+await for (final line in encodeNdjsonEnvelopes(Stream.value(envelope))) {
+  // Write line to a file or stdout
+}
+```
+
+See [example/lib/main.dart](example/lib/main.dart) and the unit tests for more examples.
+
+## Versioning
```
mpkorstanje: Please remove this section. It's all kinda obvious.

dart/analysis_options.yaml
```diff
@@ -0,0 +1,5 @@
+include: package:very_good_analysis/analysis_options.yaml
```
mpkorstanje: Why this linter specifically?

dart/CHANGELOG.md
mpkorstanje: I see. We can probably automate that somewhat in https://github.com/cucumber/polyglot-release.

dart/README.md
```diff
@@ -0,0 +1,154 @@
...
+See the repository root [CHANGELOG.md](../CHANGELOG.md) for release notes.
+
+## Development
+
+From the `dart/` directory:
+
+```sh
+make generate   # requires Ruby; regenerates lib/src/generated/ from JSON schemas
+dart pub get
+dart format .
```
mpkorstanje, line 117: Ideally the code is already formatted. Why does it need to be ran again?

dart/README.md
```diff
@@ -0,0 +1,154 @@
...
+dart analyze
+dart test
+dart pub publish --dry-run
+dart run benchmark/ndjson_benchmark.dart
+```
+
+## Performance
```
mpkorstanje: Please remove this section.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- [Dart] Add initial `cucumber_messages` package with generated message types from the Cucumber JSON schema ([#431](https://github.com/cucumber/messages/pull/431/))
+- [Dart] Add NDJSON read/write helpers (`parseEnvelopeJson`, `decodeNdjsonEnvelopes`, `encodeNdjsonEnvelopes`) with parse errors that include line number context ([#431](https://github.com/cucumber/messages/pull/431/))
+- [Dart] Add time conversion helpers between Cucumber `Timestamp`/`Duration` and Dart `DateTime`/`Duration`, truncating sub-microsecond precision ([#431](https://github.com/cucumber/messages/pull/431/))
+- [Dart] Add `IdGenerator` interface ([#431](https://github.com/cucumber/messages/pull/431/))
+- [Dart] Require all schema-required message fields in constructors and throw `SchemaViolationException` when required JSON properties are missing during deserialization ([#431](https://github.com/cucumber/messages/pull/431/))
```
mpkorstanje: You can probably summarize this a bit. 

Keep in mind that this is an end-user facing document. How the sausage is made doesn't matter that much.

dart/README.md
```diff
@@ -0,0 +1,154 @@
...
+Run a quick local throughput benchmark for envelope parsing and NDJSON read/write:
+
+```sh
+dart run benchmark/ndjson_benchmark.dart
+```
+
+This benchmark is intended for regression checks across commits, not cross-machine
+comparisons.
+
+Acceptance tests under `test/acceptance/` read fixtures from `../testdata/` in the
```
mpkorstanje, line 125: This can stay, but I think it needs a header.

.github/workflows/test-dart.yml
mpkorstanje: Also, I can't find any sort of lifecycle policies for Dart. When does a version stop being supported? Ideally, to keep CI light we only build and test against the currently supported releases.

dart/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - cucumber
+  - testing
+  - json
+  - ndjson
+
+dependencies:
+
+dev_dependencies:
+  test: ^1.31.0
+  very_good_analysis: ^8.0.0
```
mpkorstanje: I'm missing some sort of a lock file to keep the build stable. If none is available, please use exact versions instead. Renovate can probably update the dependencies.

mpkorstanje (changes requested):
Looks good. Bunch of remarks and questions below.

.github/workflows/test-dart.yml
Fried-man: The two versions are intentional:

- `3.7.0` is the package’s declared compatibility floor from `dart/pubspec.yaml` (`>=3.7.0 <4.0.0`).
- `3.12.0` is the latest stable Dart SDK ([source](https://dart.dev/resources/whats-new#3-12-release)).

Dart’s own [support policy](https://dart.dev/tools/sdk#support-policy) only supports the latest stable SDK:

> The Dart team supports only the latest, stable version of the Dart SDK. When a new major or minor version is released, older versions are no longer supported.

So I would not describe `3.7.0` as a currently supported Dart release. The reason to keep it in CI is because it verifies that our published lower bound is honest.

We need `3.7.0` specifically because this package relies on `// dart format off`, which was added in Dart 3.7 ([change blog section](https://dart.dev/blog/announcing-dart-3-7#new-formatting-style-in-the-dart-formatter)).

Dart packages commonly declare SDK ranges because consumers may be on different Dart versions, and the package manager uses that range to determine whether the package is compatible with a given project. If we keep `>=3.7.0 <4.0.0`, CI should test both the minimum declared SDK and the latest stable SDK so we validate the compatibility range we publish.

Fried-man: We could validate only the minimum but it does not give us validation that this works on the latest stable. If we wanted to, 3.12 could be replaced with a `stable` key word to point to latest version supported so this does not need to be updated manually. However, this loses determinism.

dart/benchmark/ndjson_benchmark.dart
Fried-man: Makes sense, removed.

dart/example/lib/main.dart
Fried-man: Makes sense, removed.

dart/lib/src/id_generator.dart
Fried-man: Agree, it is dead code. Removed.

dart/analysis_options.yaml
Fried-man: I chose [very_good_analysis](https://pub.dev/packages/very_good_analysis) because it is stricter than the default Dart analysis settings and catches more issues during CI before review.

That said, I am not attached to this specific lint preset. I am happy to switch back to the default Dart lints if you prefer keeping the contribution closer to vanilla dart linting.

dart/pubspec.yaml
Fried-man: That makes sense. I agree CI should be reproducible.

For Dart packages, the standard guidance is a bit different from apps. Dart says regular packages usually should not commit `pubspec.lock`, while application packages should commit it ([Dart docs](https://dart.dev/tools/pub/private-files#pubspeclock)). The reason is that packages are expected to work across a compatible dependency range while apps usually want one exact dependency graph.

If we want deterministic CI for this repository, we could track `pubspec.lock` in git. We can exclude it from the published package with `.pubignore`, since `dart pub publish` respects `.pubignore` and shows the files that will be published during `dart pub publish --dry-run` ([publishing docs](https://dart.dev/tools/pub/publishing#what-files-are-published)). This solution lets external consumers use the intended compatibility range for the package while we can still pin it exactly for CI.

dart/README.md
Fried-man: Sounds good, removed.

dart/README.md
Fried-man: Sure, removed.

dart/README.md
Fried-man: Added a header

dart/README.md
Fried-man: The code will already be formatted. However, if they are authoring new source for this package they will need to format it (via CLI or IDE). 

This section is intended for developing within the repo rather than consuming the package externally. I can remove it if you think it's unnecessary.

CHANGELOG.md
Fried-man: Sounds good, condensed it.

dart/pubspec.yaml
Fried-man: Dart’s dependency docs recommend version constraints/caret syntax for packages and note that exact versions limit adoption because they add constraints to consumers’ dependency resolution ([dependency docs](https://dart.dev/tools/pub/dependencies#traditional-syntax)).

Fried-man:
Done with changes. 

Responded to feedback, added funding link, and updated the code generation for dart generated files to include public docs from JSON schema.

dart/pubspec.yaml
mpkorstanje: > If we want deterministic CI for this repository, we could track pubspec.lock in git. We can exclude it from the published package with .pubignore

Please do that. 

Aside from deterministic CI, it also ensures that we don't immediately pull in the latest greatest compromised package when a package maintainers gets their credentials stolen again. 

Admittedly, this has mostly been a problem with NPM but the principle is the same.

.github/workflows/test-dart.yml
mpkorstanje: I'm not a fan. But it will do.

dart/analysis_options.yaml
mpkorstanje: Please switch it to the default. Fewer dependencies is better and the closer we stay to the "core" the more likely it will be that this won't need any changes in the future, or if it does that they'll be well documented.

dart/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
+
+## 0.0.1
```
mpkorstanje, line 3: Note to self: Change this with `polyglot-release`.

dart/pubspec.yaml
```diff
@@ -0,0 +1,26 @@
+name: cucumber_messages
+description: JSON schema-based messages for Cucumber inter-process communication.
+version: 0.0.1
```
mpkorstanje, line 3: Note to self: Change this in `polyglot-release`.

dart/README.md
```diff
@@ -0,0 +1,128 @@
...
+);
+
+final jsonLine = envelopeToJsonString(envelope);
+
+await for (final line in encodeNdjsonEnvelopes(Stream.value(envelope))) {
+  // Write line to a file or stdout
+}
+```
+
+See [example/lib/main.dart](example/lib/main.dart) and the unit tests for more examples.
```
mpkorstanje: These don't exist anymore right?

dart/README.md
```diff
@@ -0,0 +1,128 @@
...
+    git:
+      url: https://github.com/cucumber/messages.git
+      path: dart
+```
+
+When published to [pub.dev](https://pub.dev/packages/cucumber_messages), add:
+
+```yaml
+dependencies:
+  cucumber_messages: ^0.0.1
```
mpkorstanje: Probably best to add a more explicit place holder version here.

dart/README.md
```diff
@@ -0,0 +1,128 @@
...
+dart test
+dart pub publish --dry-run
+```
+
+## Acceptance Test Suite
+
+Acceptance tests under `test/acceptance/` read fixtures from `../testdata/` in the
+monorepo and are excluded from the published package.
+
+For the full multi-language workflow, see the repository root [README](../README.md).
```
mpkorstanje: Don't think this sentence makes sense

mpkorstanje (commented):
Alright. Last few nitpicks and we're done. Please check all unresolved discussions.

dart/analysis_options.yaml
Fried-man: Sounds good. Updated to use vanilla dart linting

dart/pubspec.yaml
Fried-man: Makes sense. Updated to track the lockfile in git and ignore for the published version

dart/README.md
Fried-man: Updated to be an explicit placeholder

dart/README.md
Fried-man: Yeah it does not exist anymore. Removed stale reference

dart/README.md
Fried-man: Updated to make more sense

Fried-man:
Done & passing on my CI

dart/pubspec.yaml
Fried-man: Also, updated the CI to use the lock file for validation

mpkorstanje (approved):

mpkorstanje:
Cheers! All good. I'll merge this once I got publishing sorted.

mpkorstanje:
It's going to be too warm for the next two weeks. I don't expect I'll get anything done soon.

mpkorstanje (approved):

PR #432, opened 2026-06-01
Remove cmate in favour of native CMake + CPM

daantimmer:
### 🤔 What's changed?

- Removed custom cmate scripts and replaced it with native cmake instead
- Use Cmake Package Manager (CPM) to (optionally) download nlohmann/json
- Update cpp.hpp.erb codegen to indent properly
- Update cpp.hpp.erb to treat all functions equal on where the return type should be located (same line instead of separate line)
- Update codegen to write to separate folders. Separating custom maintained sources/headers and generated files to make the maintenance work more streamlined. (The manually maintained sources are easily overlooked)

The custom cmate script adds no value and only confusion on maintaining the cpp codebase. Secondly its an unmaintained script by a third party.

As for the codegen, there were some inconsistencies.

I am not too sure about the separation of generated vs non-generated code. It makes reading and maintaining a lot easier. However it does mean that this is a very large change because of all the file moves.

Regarding the checklist's `#pragma once` todo item. This is a highly debated topic in C++ land. However as a public library we should follow the core guidelines as closely as possible, and thus follow the rule: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf8-use-include-guards-for-all-header-files

I can, if preferred, do this in a separate pull request.

On that note I'll also add tooling (in a separate PR) to check for compliancy with the c++ core guidelines.

### 📋 Checklist:

- [x] ~Update supporting code to maintain a consistent coding style~ **to be done in separate PR**
- [x] ~Update test-cpp to actually use different clang and gcc versions~ **to be done in separate PR**
- [x] ~Replace `#pragma once` with proper include guards.~ **to be done in separate PR**
- [x] Tooling for checking core guidelines

----


daantimmer:
I've verified these changes do not affect cucumber/gherkin -> https://github.com/cucumber/gherkin/pull/633/checks and are fully backwards compatible.

.github/workflows/codeql-cpp.yaml
```diff
@@ -27,28 +27,53 @@ concurrency:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ccache
+
+      - name: Cache CPM packages
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ env.CPM_SOURCE_CACHE }}
+          key: cpm-${{ hashFiles('cpp/cmake/dependencies.cpm.cmake') }}
+          restore-keys: cpm-
```
mpkorstanje, line 54: Reading https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#input-parameters-for-the-cache-action, I'd expect

```suggestion
          restore-keys: |
              cpm-${{ hashFiles('cpp/cmake/dependencies.cpm.cmake') }}
              cpm-
```

.github/workflows/codeql-cpp.yaml
```diff
@@ -27,28 +27,53 @@ concurrency:
...
+
+      - name: Cache ccache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ env.CCACHE_DIR }}
+          key: ccache-analysis-${{ github.ref }}
+          restore-keys: ccache-analysis-
```
mpkorstanje, line 61: Idd.

.github/workflows/test-cpp.yml
```diff
@@ -21,29 +21,102 @@ on:
...
+      - name: Install build tools (macos)
+        if: startsWith(matrix.config.os, 'macos')
+        run: brew install ccache llvm@${{ matrix.config.version }}
+
+      - name: Cache CPM packages
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: ${{ env.CPM_SOURCE_CACHE }}
+          key: cpm-${{ hashFiles('cpp/cmake/dependencies.cpm.cmake') }}
+          restore-keys: cpm-
```
mpkorstanje, line 72: Idd.

mpkorstanje (commented):
LGTM, but I don't quite understand the caching setup for Github Actions.

.github/workflows/codeql-cpp.yaml
daantimmer: Then you didn't read the docs correctly 😉 The `key` is used to first do a lookup (and used as the key to store).
If a key by that value isn't found it'll use the `restore-keys` in order from top to bottom to try and search for other keys.

Having the `key` value also in the `restore-keys` list makes the first entry for `restore-keys` redundant.

So, in short, the cache-action will do this:
1. search for `key`
2. if not found search for `restore-keys[0]`
3. if not found search for `restore-keys[1]`
4. etc

By making the `keys` one as specific as possible we always ensure we fetch the correct cache for that key. If it doesn't exist (in this case the dependencies.cpm.cmake was changed), then it'll use `cpm-` as the key to look for. Most likely finding the previous version cache of the previous version of dependencies.cpm.cmake.

PR #462, opened 2026-06-28
[cpp] Add automated tests based on testdata, simplify template usage.

daantimmer:
### 🤔 What's changed?

- Added deserialization code generation
- Added unit tests based on testdata
- Made code adhere to codestyle
- Added additional compilers to test against
- I've simplified the template usage in util.hpp to make it more maintainable

### ⚡️ What's your motivation? 

There were no tests and for that we need to be able to deserialize json objects to envelopes/messages. And the template usage was, written in a smart manner, but not really SOLID and extensible.

### 🏷️ What kind of change is this?

- :bank: Refactoring/debt/DX (improvement to code design, tooling, etc. without changing behaviour)

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)
- [x] I have added/updated tests to cover my changes.
- [x] Users should know about my change
  - [x] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/PULL_REQUEST_TEMPLATE.md)*

PR #483, opened 2026-07-14
cpp: Allow finding GTest when not fetching deps

musicinmybrain:
### 🤔 What's changed?

This is the same idea as https://github.com/cucumber/messages/pull/478, extended to find `GTest` targets needed for C++ tests since https://github.com/cucumber/messages/pull/462. (I’m happy to see an actual usable C++ test suite, by the way!)

Syntax-wise, this should be compatible with the declared CMake minimum version

https://github.com/cucumber/messages/blob/d28e52ca27b280a583a5c5799de15672ed712d58/cpp/CMakeLists.txt#L18

but https://cmake.org/cmake/help/latest/module/FindGTest.html#imported-targets indicates that CMake 3.23 or later is necessary in order for the `GTest::gtest`, `GTest::gtest_main`, `GTest::gmock`, and `GTest::gmock_main` to actually be defined based on system packages.

Straightforwardly build and run the tests in fully-offline environments, particularly Linux distribution packaging.

N/A

### 📋 Checklist:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Fixed
+- [cpp] Allow building tests with system GTest when fetching dependencies is disabled. ([#483](https://github.com/cucumber/messages/pull/483))
 
 ## [34.0.2] - 2026-07-13
 ### Added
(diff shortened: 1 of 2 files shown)


daantimmer (approved):

daantimmer:
@mpkorstanje lgtm. Didn't expect tests needing to be build when not directly developing for this repo. 

But it doesn't break anything. So completely OK to add.

musicinmybrain:
> @mpkorstanje lgtm. Didn't expect tests needing to be build when not directly developing for this repo.
> 
> But it doesn't break anything. So completely OK to add.

Thanks for the review!

For context, we run test suites whenever we can in Fedora’s packages. It adds confidence that we’ve built things correctly, it helps reveal bugs on less popular architectures (we have `x86_64`, `aarch64`, `ppc64le`, and `s390x` as primary architectures), it helps us catch unexpected behavior changes from dependency updates, and it even occasionally helps reveal things like compiler bugs in new versions of GCC.

mpkorstanje:
@musicinmybrain makes total sense.

PR #484, opened 2026-07-14
cpp: Update google/googletest with Renovate

mpkorstanje:
### 📋 Checklist:


cpp/cmake/dependencies.cpm.cmake
```diff
@@ -48,7 +50,7 @@ if(CUCUMBER_MESSAGES_FETCH_DEPS)
         CPMAddPackage(
             NAME googletest
             GITHUB_REPOSITORY google/googletest
-            GIT_TAG 52eb8108c5bdec04579160ae17225d66034bd723 # v1.17.0
+            GIT_TAG v${GOOGLE_TEST_VERSION}
```
mpkorstanje, line 53: @daantimmer I assume we don't need the hash. If we do, I can rework this.

PR #485, opened 2026-07-14
[cpp] add git-hash to dependencies

daantimmer:
### 🤔 What's changed?

CPM's AddPackage now uses the git hash of nlohmann/json and google/googletest instead of their version number.

In general it is better to pin to hash than a git tag since git tags are mutable and hashes are not.

### 📋 Checklist:

----

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -18,9 +18,6 @@ endif()
 cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 project(cucumber_messages VERSION ${VERSION} LANGUAGES CXX)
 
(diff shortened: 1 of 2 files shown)


daantimmer:
@mpkorstanje can I ask you to do your renovate magic?

if it suits renovate better it is also possible to write the HASH_VERSION list as:
```cmake
set(NLOHMANN_JSON_HASH_VERSION "55f93686c01528224f448c19128836e7df245f72;3.12.0")
```

instead of

```cmake
set(NLOHMANN_JSON_HASH_VERSION 55f93686c01528224f448c19128836e7df245f72 3.12.0)
```

And could the same be done for the CPM_DOWNLOAD_VERSION?

PR #474, opened 2026-07-04
[dart]  Add equals, hashCode and toString to generated messages

Fried-man:
# Add value semantics (== hashCode toString) to generated Dart message classes

### 🤔 What's changed?

Generated Dart message classes now have value semantics. The generator emits three members per class. `operator ==` gives structural equality where scalar `$ref` and enum fields use `==` and `List` fields use element-wise `_listEquals`. `hashCode` uses `Object.hashAll([...])` with list fields hashed by content. `hashAll` handles any field count including the 21-field `Envelope`. `toString` returns `ClassName{field: value ...}`.

Changes live in the templates so all ~61 types get it at once. `codegen/templates/dart.dart.erb` adds the four members. `codegen/templates/dart.index.dart.erb` adds a private `_listEquals` helper with zero new dependencies (no `package:collection`). Regenerated all `*.g.dart` files plus `messages.dart` and added `dart/test/src/value_semantics_test.dart`.

### ⚡️ What's your motivation?

Generated classes were immutable data holders with no equality hash readable `toString`. So `a == b` was identity comparison. Structurally-equal messages compared unequal and could not be Set/Map keys or deduplicated. `print(envelope)` gave `Instance of 'Envelope'`. This was the biggest divergence from the Java implementation which generates `equals`/`hashCode`/`toString` and from Dart conventions.

### 🏷️ What kind of change is this?

- :zap: New feature (non-breaking change which adds new behaviour)

### ♻️ Anything particular you want feedback on?

n/a

### 📋 Checklist:

- [x] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)
- [x] I've changed the behaviour of the code
  - [x] I have added/updated tests to cover my changes.
- [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.
- [x] Users should know about my change
  - [x] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

PR #498, opened 2026-07-21
dart: Fix linter issues

Fried-man:
### 🤔 What's changed?

I removed an unneeded null check from code generated lists, unnecessary braces from `toString` for the code generated objects, and renamed `messages.dart` to `messages.g.dart` per dart naming convention for code generated files. Also, I removed the exclude from the `analysis_options.yaml` so the CI will catch this static analysis regression from happening again.

So [pana](https://pub.dev/packages/pana) is what is used to grade packages published on pub.dev out of 160 points. For whatever reason, they run the linter on the entire `lib/` ignoring the part (code-generated code) that I told the linter to exclude in the `analysis_options.yaml`. This behavior was exposed from the changes introduced in #474 which decreased the pana score by [20 points on pub.dev](https://pub.dev/packages/cucumber_messages/versions/34.2.0/score). In this case, it was reasonable lint catches so I went ahead and updated the code generator to fix them. Pana is now reporting a perfect score again. 

Additionally, `messages.dart` was updated to have a `.g.dart` suffix since it is code generated.

### 📋 Checklist:

----


mpkorstanje (commented):
LGTM, but please don't use title case for your PR descriptions. We squash the commits so the PR title becomes the git commit subject. You can follow the best practices for those (https://cbea.ms/git-commit/).

